PR #1, opened 2025-08-20
Initial setup

MartijnBogaert:
LBRON-359

A basic microservice stack has been set up. These are the services involved:
- Identifier
- Dispatcher
- SPARQL parser
- Virtuoso triplestore
- Resources
- File service

For now, resources of type `eli:LegalExpression` have been made publicly accessible through the endpoint `/legal-expressions`.

> Alongside the repo setup, a *Decide* team has also been made, with all members having admin privileges to the repo.

### DIFF
--- a/config/virtuoso/virtuoso.ini
+++ b/config/virtuoso/virtuoso.ini
@@ -1,0 +1,262 @@
+;
+;  virtuoso.ini
+;
+;  Configuration file for the OpenLink Virtuoso VDBMS Server
+;
+;  To learn more about this product, or any other product in our
+;  portfolio, please check out our web site at:
+;
+;      http://virtuoso.openlinksw.com/
+;
+;  or contact us at:
+;
+;      general.information@openlinksw.com
+;
+;  If you have any technical questions, please contact our support
+;  staff at:
+;
+;      technical.support@openlinksw.com
+;
+
+;
+;  Database setup
+;
+[Database]
+DatabaseFile			= /usr/local/virtuoso-opensource/var/lib/virtuoso/db/virtuoso.db
+ErrorLogFile			= /usr/local/virtuoso-opensource/var/lib/virtuoso/db/virtuoso.log
+LockFile			= /usr/local/virtuoso-opensource/var/lib/virtuoso/db/virtuoso.lck
+TransactionFile			= /usr/local/virtuoso-opensource/var/lib/virtuoso/db/virtuoso.trx
+xa_persistent_file		= /usr/local/virtuoso-opensource/var/lib/virtuoso/db/virtuoso.pxa
+ErrorLogLevel			= 7
+FileExtend			= 200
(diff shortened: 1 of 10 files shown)


config/virtuoso/virtuoso.ini
```diff
@@ -0,0 +1,262 @@
...
+;NumberOfBuffers          = 4000000
+;MaxDirtyBuffers          = 3000000
+;; Uncomment next two lines if there is 64 GB system memory free
+;NumberOfBuffers          = 5450000
+;MaxDirtyBuffers          = 4000000
+;;
+;; Note the default settings will take very little memory
+;; but will not result in very good performance
+;;
+NumberOfBuffers          = 10000
```
Rahien, line 116: this is super low, we should probably use something a little higher already, like the 4GB version or so would alsoo increase vector size, max vector size and maxquerymem.
ofc these settings will change (and likely increase further) during the project

config/virtuoso/virtuoso.ini
```diff
@@ -0,0 +1,262 @@
...
+DynamicLocal			= 0
+DefaultHost			= localhost:8890
+
+
+[SPARQL]
+;ExternalQuerySource		= 1
+;ExternalXsltSource 		= 1
+;DefaultGraph      		= http://localhost:8890/dataspace
+;ImmutableGraphs    		= http://localhost:8890/dataspace
+ResultSetMaxRows           	= 10000
```
Rahien, line 239: lets put this one higher too, same for maxconstruct triples. I would remove the max query cost estimation time and set max to something a lot higher like 600?

docker-compose.yml
```diff
@@ -0,0 +1,58 @@
...
+      - ./data/authorization:/data
+    labels:
+      - "logging=true"
+    restart: always
+    logging: *default-logging
+  virtuoso:
+    image: redpencil/virtuoso:1.3.0
+    environment:
+      SPARQL_UPDATE: "true"
+      DEFAULT_GRAPH: "http://mu.semte.ch/application"
```
Rahien: not sure if there is any benefit to putting this. i would leave it out

Rahien: reason: the application graph often doesn't exist anymore, public would be a better default. Then again, I find it better to NOT set a default graph because it will give you results you don't expect if you expect to be querying on all graphs instead

docker-compose.yml
```diff
@@ -0,0 +1,58 @@
...
+    volumes:
+      - ./config/virtuoso/virtuoso.ini:/data/virtuoso.ini
```
Rahien, line 36: is it worth already creating a virtuoso.prod.ini and putting the development version in our docker-compose.dev.yml?

Rahien (approved):

docker-compose.yml
MartijnBogaert: Good to know! I left it out.

PR #2, opened 2025-09-03
feat: add project script to transform ttl to n-triples

mirdono:
This adds a project script to the add that allows to convert turtle files into their corresponding n-triples files.

This script is to be used as part of a workaround for the [odrl-parser](https://github.com/lblod/odrl-parser-service) service[^1] to allow it to process files containing ODRL policies. Ideally that service would directly read from turtle files. But the service is written in common lisp[^2] and there is currently no turtle parser available in that language. In anticipation of a turtle parser, we opted to read ODRL policies as n-triples, for which there is a parser available. To avoid having to actually write policies as n-triples this script can be used to automatically convert ODRL policies in turtle to n-triples.

The script will look for turtle files in `config/odrl-parser/`[^3] and write the generated n-triples files to the same directory. The file names of the generated files are taken from the source files.

## How to test
Run the script with the following command[^4]:

``` shell
mu script project-scripts ttl-to-ntriples
```

By default, the script does *not* overwrite already existing files, it simply skips turtle files for which there already exists an n-triples file with the same name. This behaviour can be changed by providing the `-r` (or `--replace`) flag to the script to instruct it to overwrite any existing files:

``` shell
mu script project-scripts ttl-to-ntriples --replace
```

## Notes
- The `CONFIG_PATH` from which input files are read is currently hardcoded. Should this script be used in other contexts this can be made configurable.
- The replace behaviour is intentionally kept as simple as possible: simply overwrite any existing file with the same name. It will not ask for confirmation, allow to append instead, or other fancy stuff.

## Related tickets
- LBRON-328

[^1]: A service, currently under development, that translates ODRL policies into authorisation configurations for [sparql-parser](https://github.com/mu-semtech/sparql-parser).

[^2]: Common lisp was choosen to facilitate future integration with sparql-parser.

[^3]: This path is currently hardcoded and is the location from where the odrl-parser service reads its ODRL policies from.

[^4]: Here `mu` refers to [mu-cli](https://github.com/mu-semtech/mu-cli) which should be installed on your machine for this to work.

### DIFF
--- a/config/odrl-parser/config.ttl
+++ b/config/odrl-parser/config.ttl
@@ -1,0 +1,312 @@
+# ODRL policy in ttl format corresponding. This policy matches the simplified
+# LMB authorisation policy in the lisp file.
+@prefix ext: <http://mu.semte.ch/vocabularies/ext/>.
+@prefix odrl: <http://www.w3.org/ns/odrl/2/>.
+@prefix dct: <http://purl.org/dc/terms/>.
+@prefix vcard: <http://www.w3.org/2006/vcard/ns#>.
+@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>.
+@prefix mandaat: <http://data.vlaanderen.be/ns/mandaat#>.
+@prefix skos: <http://www.w3.org/2004/02/skos/core#>.
+@prefix sh: <http://www.w3.org/ns/shacl#>.
+@prefix lmb: <http://lblod.data.gift/vocabularies/lmb/>.
+@prefix besluit: <http://data.vlaanderen.be/ns/besluit#>.
+@prefix regorg: <http://www.w3.org/ns/regorg#>.
+@prefix persoon: <http://data.vlaanderen.be/ns/persoon#>.
+
+# TODO: Extend with extra `ext:*` elements?
+ext:muAuthProfile a odrl:Profile ;
+  dct:description "ODRL profile describing access to mu-auth resources." .
+
+ext:examplePolicyLMB a odrl:Set ;
+  odrl:profile ext:muAuthProfile ;
+  odrl:permission ext:allowReadForPublic ,
+    ext:allowReadForMandatarisOrg ,
+    ext:allowWriteForMandatarisOrg ,
+    ext:allowReadForVendorOnMandatarisOrg ,
+    ext:allowReadForPoliceCouncilOnMandatarisOrg ,
+    ext:allowReadForAuthenticatedOnViewOnlyModules .
+
+# parties
+ext:lmbSystem a odrl:Party ;
+  vcard:fn "LMB System" ;
+  # no query here so no users match
+  # TODO: Use sparql-parser's 'NEVER' constraint to ensure group is not used in
+  # configuration?
+  dct:description "The LMB system party used as an assigner of the permission in this profile" .
+
+ext:publicParty a odrl:PartyCollection ;
+  vcard:fn "Public user" ;
+  # NOTE (20/08/2025): Ideally it should not be necessary to add this as it will
+  # not be used in the sparql-parser configuration. Is it necessary to
+  # differentiate between this case, which is "all users", and the above system
+  # party, which is "no users"?
+  #
+  # this query matches every session and has no parameters
+  ext:definedBy """
+  PREFIX ext: <http://mu.semte.ch/vocabularies/ext/>
+  PREFIX muAccount: <http://mu.semte.ch/vocabularies/account/>
+  PREFIX mu: <http://mu.semte.ch/vocabularies/core/>
+  SELECT DISTINCT * WHERE {
+    VALUES ?session {
+      <SESSION_ID>
+    }
+    ?session a ?thing .
+  }
+  """ ;
+  dct:description "This party represent all (possibly not logged in) users of the system." .
+
+ext:authenticatedParty a odrl:PartyCollection ;
+  vcard:fn "Authenticated user" ;
+  # no `ext:queryParameters` as they will not be used for this party
+  ext:definedBy """
+  PREFIX ext: <http://mu.semte.ch/vocabularies/ext/>
+  PREFIX mu: <http://mu.semte.ch/vocabularies/core/>
+  SELECT DISTINCT ?session_group ?session_role WHERE {
+    <SESSION_ID> ext:sessionGroup/mu:uuid ?session_group ;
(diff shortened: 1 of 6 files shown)


config/odrl-parser/config.ttl
```diff
@@ -0,0 +1,308 @@
...
+  dct:description "This party represent all (possibly not logged in) users of the system." .
+
+ext:authenticatedParty a odrl:PartyCollection ;
+  vcard:fn "Authenticated user" ;
+  # no `ext:queryParameters` as they will not be used for this party
+  ext:definedBy """
+  PREFIX ext: <http://mu.semte.ch/vocabularies/ext/>
+  PREFIX mu: <http://mu.semte.ch/vocabularies/core/>
+  SELECT DISTINCT ?session_group ?session_role WHERE {
+    <SESSION_ID> ext:sessionGroup/mu:uuid ?session_group ;
```
Rahien, line 65: this requires the users to have at least one group though. i know that currently this is how all login services work, but maybe we should only require the role for 'logged in' users? wdyt?

config/odrl-parser/config.ttl
```diff
@@ -0,0 +1,308 @@
...
+      }
+    }}
+  }
+  """ ;
+  ext:queryParameters "session_group", "session_role" ;
+  dct:description "This party collection represents all users who can access the information regarding communes or OCMWs through their vendor api key" .
+
+ext:policeCouncilParty a odrl:PartyCollection ;
+  vcard:fn "Politieraad lezer" ;
+  ext:definedBy """
```
Rahien, line 116: i'm guessing at some point we will be defining a proper prefix for these things, but for now ext makes sense!

config/odrl-parser/config.ttl
```diff
@@ -0,0 +1,308 @@
...
+    FILTER( ?session_role = \"LoketLB-mandaatGebruiker\" )
+  }
+  """ ;
+  ext:queryParameters "session_group", "session_role" ;
+  dct:description "This party collection represents all users who can access the information regarding police council mandates. This is defined by the members of communes that share the control of this police council." .
+
+
+# graphs
+ext:lmbPublicSlice a odrl:AssetCollection ;
+# TODO: Add name that should be used for define-graph function
```
Rahien: any reason to wait?

config/odrl-parser/config.ttl
```diff
@@ -0,0 +1,308 @@
...
+ext:lmbAuthenticatedPublicSlice a odrl:AssetCollection ;
+# TODO: Add name that should be used for define-graph function
+  ext:graphPrefix <http://mu.semte.ch/graphs/authenticated/public> ;
+  ext:shaclShape ext:lmbAuthenticatedPublicShape ;
+  dct:description "This asset collection contains all information that is available to all authenticated users in the context of the LMB app." .
+
+ext:lmbOrganizationMandatesSlice a odrl:AssetCollection ;
+# TODO: Add name that should be used for define-graph function
+  ext:graphPrefix <http://mu.semte.ch/graphs/organizations/> ;
+  # TODO: Can we avoid having to specify these parameters again?
```
Rahien, line 148: hmmm 🤔 in theory you can infer from where the slice is used, but it makes things a bit complicated and risks misconfiguration that is hard to track, this way sparql-parser can in theory (not sure if it does) error is there is a mismatch between the slice and the partycollection. i would keep it for now

scripts/ttl-to-ntriples/ttl-to-ntriples.py
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/python
+
+import os
+import argparse
+from rdflib import Graph
+
+# TODO: make configurable
```
Rahien: i think this one is done \o/

scripts/ttl-to-ntriples/ttl-to-ntriples.py
```diff
@@ -0,0 +1,27 @@
...
+
+if os.path.isdir(CONFIG_PATH):
+    os.chdir(CONFIG_PATH)
+    for file in os.listdir():
+        target = file.replace(TTL_EXT, NTRIPLES_EXT)
+        if file.endswith(TTL_EXT) and (args.replace or not os.path.exists(target)):
+            print(" >> Found " + file)
+            graph = Graph()
+            graph.parse(file, format="turtle")
+            graph.serialize(format="nt11", destination=target)
```
Rahien, line 26: would it make sense to do a better reification of the blank nodes than use the _ prefix? Will the odrl-parser be able to handle blank nodes properly?

Rahien (approved):
Looks good, left some comments! I mostly would like to get your view on the way blank nodes are dealt with

scripts/ttl-to-ntriples/ttl-to-ntriples.py
mirdono: The odrl-parser service itself can so far handle this. I did have to work around virtuoso's `INSERT DATA` not allowing blank nodes with the `_:` prefix.  It is probably cleaner to use rdflib's skolemisation functionality here to avoid that workaround. Pushed a new commit with a added skolemisation step.

scripts/ttl-to-ntriples/ttl-to-ntriples.py
mirdono: What I meant here was make to make it configurable by allowing a path to be passed as argument to the script. I will clarify the TODO note.

config/odrl-parser/config.ttl
mirdono: No specific reason, in the back of my head I was still mulling over whether it is necessary or we can always derive a name from already information. But at least allowing users to explicitly specify a name for graphs/asset collections seems like a good thing. (Can still make it "optional" by having the odrl-parser come with a name if none is provided.) I will add names.

config/odrl-parser/config.ttl
mirdono: Indeed, but I prefer to have a working poc for that generating sparql-parser configurations from this actually works as expected before going hunting for more appropriate/permanent prefixes.

config/odrl-parser/config.ttl
mirdono: I am not sure I am following here. The (part of the) highlighted query is part of the configuration provided by whoever defines the authorisation policy. If the relations between users and their groups/roles is different a different query would have to be defined, same as you would have to do for the configuration of sparql-parser.

PR #3, opened 2025-09-05
Implement ELI(-DL) classes and properties targeted by OParl and OSLO mappings

MartijnBogaert:
OPH-404

All classes and properties that appear as results in the (theoretical) mappings from OParl and OSLO to ELI(-DL) have been added to the resource definitions.

Mappings:
- https://app.gitbook.com/o/-MP9Yduzf5xu7wIebqPG/s/grRYjcouX0uZY2P6PYBl/data-analysis-research/map-oparl-to-eli-dl
- https://app.gitbook.com/o/-MP9Yduzf5xu7wIebqPG/s/grRYjcouX0uZY2P6PYBl/data-analysis-research/map-oslo-to-eli-dl

> The OParl to ELI(-DL) mapping is still a work in progress, so some target RDF classes/properties are still *empty*. To circumvent this, the `oparl-temp` prefix is used temporarily for those, pending mapping completion.

All introduced resources have been made publicly availble through mu-auth and dispatcher.


MartijnBogaert:
@brechtvdv I followed the mappings as closely as possible, but I can imagine some properties not really being necessary for our case. E.g. I'm thinking of this set of properties that appears essentially for all resources in the OParl mapping:
```
(:license :string ,(s-prefix "oparl-temp:license"))
(:keyword :string ,(s-prefix "oparl-temp:keyword"))
(:web :url ,(s-prefix "oparl-temp:web"))
(:deleted :boolean ,(s-prefix "oparl-temp:deleted"))
```
If you would have some time to go through the PR, it would in any case seem like a good idea to have a look at the classes/properties that start with the `oparl-temp` prefix. Maybe some of those properties can go away, or maybe you know already what the correct RDF classes/properties should be. Thanks!

config/resources/eli-dl.lisp
```diff
@@ -0,0 +1,138 @@
...
+                    :as "documented-by-a-realization-of")
+              (work :via ,(s-prefix "eli-dl:recorded_in_a_realization_of")
+                    :as "recorded-in-a-realization-of")
+              (work :via ,(s-prefix "eli-dl:based_on_a_realization_of")
+                    :as "based-on-a-realization-of")
+              (foreseenActivity :via ,(s-prefix "eli-dl:consists_of")
+                                :as "consists-of")
+              (foreseenActivity :via ,(s-prefix "eli-dl:executed")
+                                :as "executed")
+              (participant :via ,(s-prefix "eli-dl:had_participation")
```
Rahien: this one doesn't seem to exist. an easy way of finding these issues is to run `mu project doc`, it will either spit out a model.ttl + plantuml file OR it will throw errors pointing to issues (in a very lisp way but still)

Rahien (changes requested):

config/resources/eli-dl.lisp
MartijnBogaert: Thanks for pointing out! Was a typo: <del>participant</del> --> participation

Rahien (approved):

PR #7, opened 2025-10-02
Add DCAT resources config

niels-triply:
Also activates corresponding sparql-parser & dispatcher config. Closes LBRON-495

Note: For the `format` resource, there is a `resource-base` of `http://example.com/formats` as well as a `birth-year` property. This was present in the DCAT config in the past. Should this be removed?


config/authorization/config.lisp
```diff
@@ -18,6 +18,9 @@
   :eli "http://data.europa.eu/eli/ontology#"
   :eli-dl "http://data.europa.eu/eli/eli-draft-legislation-ontology#"
   :oparl-temp "http://mu.semte.ch/vocabularies/ext/oparl/"
+  :skos "http://www.w3.org/2004/02/skos/core#"
+  :dcat "http://www.w3.org/ns/dcat#"
+  :dct "http://purl.org/dc/terms/"
```
mirdono: **Comment:** The `dct` prefix is defined twice.

The `dct` was already defined, third in the list, no need to add it again.

config/resources/dcat.lisp
```diff
@@ -0,0 +1,155 @@
...
+  :properties `((:name :string ,(s-prefix "rdfs:label"))
+                (:labels :language-string-set ,(s-prefix "dct:description"))
+                (:birth-year :g-year ,(s-url "http://example.com/birthYear")))
+  :has-many `((distribution :via ,(s-prefix "dct:format")
+                            :inverse t
+                            :as "distributions"))
+  :has-one `((page :via ,(s-prefix "cms:page")
+                   :as "page"))
+  :on-path "formats"
+  :resource-base (s-url "http://example.com/formats"))
```
mirdono: **Comment:** Might use a less generic resource base and add a trailing "/".

I would suggest to use a more specific resource base, as you did for the other resources. At least I would add a trailing "/" to the prefix URI so we do not end up with incorrectly prefixed resource URIs.

config/resources/dcat.lisp
```diff
@@ -0,0 +1,155 @@
...
+
+(around (:show page) (&rest args)
+  (break "This is page showing with ~A" args)
+  (let ((response (yield)))
+    (break "The response should be ~A" (jsown:to-json response))
+    (jsown:new-js ("ok" t))))
+
+(after (:show page) (&rest args)
+  (declare (ignore args))
+  (jsown:new-js ("ok" :false)))
```
mirdono: **Question:** What is the purpose of the `around` and `after` calls?

At first sight they seem to be useful for debugging or something. I am not sure we need them here.

config/resources/dcat.lisp
```diff
@@ -0,0 +1,155 @@
...
+
+(define-resource format ()
+  :class (s-prefix "dct:MediaTypeOrExtent")
+  :properties `((:name :string ,(s-prefix "rdfs:label"))
+                (:labels :language-string-set ,(s-prefix "dct:description"))
+                (:birth-year :g-year ,(s-url "http://example.com/birthYear")))
+  :has-many `((distribution :via ,(s-prefix "dct:format")
+                            :inverse t
+                            :as "distributions"))
+  :has-one `((page :via ,(s-prefix "cms:page")
```
mirdono, line 128: **Issue:** The `cms:` prefix is not yet defined.

The `cms:` prefix is not yet defined in `repository.lisp`. You can add the following to the file (assuming this is the same cms as used in the sparql-parser config):

``` common-lisp
(add-prefix "cms" "http://mu.semte.ch/vocabulary/cms/")
```

config/resources/dcat.lisp
```diff
@@ -0,0 +1,155 @@
...
+  (break "This is page showing with ~A" args)
+  (let ((response (yield)))
+    (break "The response should be ~A" (jsown:to-json response))
+    (jsown:new-js ("ok" t))))
+
+(after (:show page) (&rest args)
+  (declare (ignore args))
+  (jsown:new-js ("ok" :false)))
+
+(define-resource publisher ()
```
mirdono, line 140: **Issue:** There is no dispatcher rule for this resource type.

The dispatcher configuration does not contain a rule for this resource type, making them inaccessible. Is this intentional? If so, why include them here at all?

mirdono (changes requested):
Some minor issues and questions, see the inline comments for details.

config/authorization/config.lisp
niels-triply: Oops, you're right. I've removed it and, to prevent this thing in the future, sorted all prefixes. In the process, found another one that was already there (`org`). Also removed that one.

config/resources/dcat.lisp
niels-triply: Yes, I just removed the `birthYear` property expected on formats (that didn't seem to make any sense) and changed the resource-base. Am I correct in suspecting that the `resource-base` for the `format` resource does not really matter, because the `resource-base` is for creating new resources, and formats are ideally not created by us but taken from a value list?

config/resources/dcat.lisp
niels-triply: They were there in the original configuration and I did not remove them because I did not fully understand their purpose. I have removed them.

niels-triply:
@mirdono Thanks for the review. I've made all changes. One is left marked unresolved because I want to check an assumption, but if you see no issue, you can resolve and merge.

config/resources/dcat.lisp
mirdono: Indeed, AFAIK the resource-base is used when creating a new resource via the mu-cl-resources service. Not sure where the possible formats would come from, it does make sense we would reuse some existing ones and not invent/create our own.

mirdono (approved):
Looks good now. If you do not mind, could you squash the commits a bit together? Imo intermediary steps, for example adding something only to remove it again in a later commit, are just part of the development process and should not end up in the main history.

niels-triply:
I have squashed this into two commits (import + adaptation). Should be ready. (And thanks!)

PR #9, opened 2025-10-03
fix[resources]: use consistent prefix for dc terms

mirdono:
The resources configuration used both `dct:` and `dcterms:` as prefix, where
only the former is actually defined in `repository.lisp`. This changes all uses
of `dcterms:` to `dct:` so we use a single prefix for this spec.


MartijnBogaert (approved):
Thanks for noticing!

PR #11, opened 2025-10-16
Add mock-login backend

niels-triply:
Ref: LBRON-496

### DIFF
--- a/config/migrations/20240000000005-add-mock-users.sparql
+++ b/config/migrations/20240000000005-add-mock-users.sparql
@@ -49,7 +49,7 @@ WHERE {
 
     VALUES (?g) {
       (<http://mu.semte.ch/graphs/public>)
-      (<http://mu.semte.ch/graphs/organizations/7d7122bd82f165e091a36c830d041b59aa48e4be9ea0aa06dc4c67ef381b7505/LoketLB-mandaatGebruiker>)
+      (<http://mu.semte.ch/graphs/organizations/7d7122bd82f165e091a36c830d041b59aa48e4be9ea0aa06dc4c67ef381b7505>)
     }
     BIND("Jane" as ?firstName)
     BIND("Doe" as ?familyName)
--- a/config/migrations/20240000000005-add-mock-users.sparql
+++ b/config/migrations/20240000000005-add-mock-users.sparql
@@ -1,0 +1,61 @@
+PREFIX besluit: <http://data.vlaanderen.be/ns/besluit#>
+PREFIX skos: <http://www.w3.org/2004/02/skos/core#>
+PREFIX foaf: <http://xmlns.com/foaf/0.1/>
+PREFIX mu: <http://mu.semte.ch/vocabularies/core/>
+PREFIX ext: <http://mu.semte.ch/vocabularies/ext/>
+
+INSERT {
+
+  GRAPH ?g {
+    ?persoon a foaf:Person;
+        mu:uuid ?uuidPersoon;
+        foaf:member <http://data.lblod.info/id/bestuurseenheden/7d7122bd82f165e091a36c830d041b59aa48e4be9ea0aa06dc4c67ef381b7505>;
+        foaf:firstName ?firstName;
+        foaf:familyName ?familyName;
+        foaf:account ?account.
+    ?account a foaf:OnlineAccount;
+        mu:uuid ?uuidAccount;
+        ext:sessionGroup <http://data.lblod.info/id/bestuurseenheden/7d7122bd82f165e091a36c830d041b59aa48e4be9ea0aa06dc4c67ef381b7505>;
+        foaf:accountServiceHomepage <https://github.com/lblod/mock-login-service>;
+        ext:sessionRole "Decide-gebruiker".
+
+<http://data.lblod.info/id/bestuurseenheden/7d7122bd82f165e091a36c830d041b59aa48e4be9ea0aa06dc4c67ef381b7505>	<http://www.w3.org/ns/org#classification>	<http://data.vlaanderen.be/id/concept/BestuurseenheidClassificatieCode/5ab0e9b8a3b2ca7c5e000001> .
+
+<http://data.lblod.info/id/bestuurseenheden/7d7122bd82f165e091a36c830d041b59aa48e4be9ea0aa06dc4c67ef381b7505>	<http://data.vlaanderen.be/ns/besluit#classificatie>	<http://data.vlaanderen.be/id/concept/BestuurseenheidClassificatieCode/5ab0e9b8a3b2ca7c5e000001> .
+
+<http://data.lblod.info/id/bestuurseenheden/7d7122bd82f165e091a36c830d041b59aa48e4be9ea0aa06dc4c67ef381b7505>	<http://www.w3.org/ns/org#hasPrimarySite>	<http://data.lblod.info/id/vestigingen/f34a68fdd29a73b4bb1e394bfedb670a> .
+
+<http://data.lblod.info/id/bestuurseenheden/7d7122bd82f165e091a36c830d041b59aa48e4be9ea0aa06dc4c67ef381b7505>	<http://purl.org/dc/terms/identifier>	"0207445287" .
+<http://data.lblod.info/id/bestuurseenheden/7d7122bd82f165e091a36c830d041b59aa48e4be9ea0aa06dc4c67ef381b7505>	<http://www.w3.org/ns/adms#identifier>	<http://data.lblod.info/id/identificatoren/a131fbc9-a78a-4c03-8dd3-bcd480c1810a> .
+
+<http://data.lblod.info/id/bestuurseenheden/7d7122bd82f165e091a36c830d041b59aa48e4be9ea0aa06dc4c67ef381b7505>	<http://www.w3.org/ns/regorg#legalName>	"Buggenhout" .
+<http://data.lblod.info/id/bestuurseenheden/7d7122bd82f165e091a36c830d041b59aa48e4be9ea0aa06dc4c67ef381b7505>	<http://purl.org/dc/terms/identifier>	"OVO002038" .
+
+<http://data.lblod.info/id/bestuurseenheden/7d7122bd82f165e091a36c830d041b59aa48e4be9ea0aa06dc4c67ef381b7505>	<http://mu.semte.ch/vocabularies/core/uuid>	"7d7122bd82f165e091a36c830d041b59aa48e4be9ea0aa06dc4c67ef381b7505" .
+<http://data.lblod.info/id/bestuurseenheden/7d7122bd82f165e091a36c830d041b59aa48e4be9ea0aa06dc4c67ef381b7505>	<http://www.w3.org/2004/02/skos/core#prefLabel>	"Buggenhout" .
+<http://data.lblod.info/id/bestuurseenheden/7d7122bd82f165e091a36c830d041b59aa48e4be9ea0aa06dc4c67ef381b7505>	<http://mu.semte.ch/vocabularies/ext/kbonummer>	"0207445287" .
+<http://data.lblod.info/id/bestuurseenheden/7d7122bd82f165e091a36c830d041b59aa48e4be9ea0aa06dc4c67ef381b7505>	<http://data.vlaanderen.be/ns/besluit#werkingsgebied>	<http://data.lblod.info/id/werkingsgebieden/6c6dcdc5ea2a8eb3d951b9149e7979afa4c7f86143a7576ce9c480e33f26cd7d> .
+
+<http://data.lblod.info/id/bestuurseenheden/7d7122bd82f165e091a36c830d041b59aa48e4be9ea0aa06dc4c67ef381b7505>	<http://www.w3.org/1999/02/22-rdf-syntax-ns#type>	<http://data.vlaanderen.be/ns/besluit#Bestuurseenheid> .
+
+ <http://data.lblod.info/id/identificatoren/a131fbc9-a78a-4c03-8dd3-bcd480c1810a>	<https://data.vlaanderen.be/ns/generiek#gestructureerdeIdentificator>	<http://data.lblod.info/id/gestructureerdeIdentificatoren/2a7c78a9-8598-4741-b1eb-cae7084f1b9b> .
+ <http://data.lblod.info/id/identificatoren/a131fbc9-a78a-4c03-8dd3-bcd480c1810a>	<http://www.w3.org/1999/02/22-rdf-syntax-ns#type>	<http://www.w3.org/ns/adms#Identifier> .
+ <http://data.lblod.info/id/identificatoren/a131fbc9-a78a-4c03-8dd3-bcd480c1810a>	<http://mu.semte.ch/vocabularies/core/uuid>	"a131fbc9-a78a-4c03-8dd3-bcd480c1810a" .
+ <http://data.lblod.info/id/identificatoren/a131fbc9-a78a-4c03-8dd3-bcd480c1810a>	<http://www.w3.org/2004/02/skos/core#notation>	"SharePoint identificator" .
+
+  }
+}
+WHERE {
+
+    VALUES (?g) {
+      (<http://mu.semte.ch/graphs/public>)
+      (<http://mu.semte.ch/graphs/organizations/7d7122bd82f165e091a36c830d041b59aa48e4be9ea0aa06dc4c67ef381b7505/LoketLB-mandaatGebruiker>)
+    }
+    BIND("Jane" as ?firstName)
+    BIND("Doe" as ?familyName)
+    BIND(CONCAT(?familyName, ?firstName) as ?volledigeNaam)
+    BIND(MD5(?volledigeNaam) as ?uuidPersoon)
+    BIND(MD5(CONCAT(?volledigeNaam, "ACCOUNT")) as ?uuidAccount)
+    BIND(IRI(CONCAT("http://data.lblod.info/id/persoon/", ?uuidPersoon)) AS ?persoon)
+    BIND(IRI(CONCAT("http://data.lblod.info/id/account/", ?uuidAccount)) AS ?account)
+}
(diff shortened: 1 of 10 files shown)


niels-triply:
Frontend at https://github.com/lblod/frontend-decide/pull/1

docker-compose.dev.yml
```diff
@@ -14,4 +14,6 @@ services:
   resource:
     restart: "no"
   file:
```
JonasVanHoof, line 16: Could you also add the migration service to the compose or compose-dev file? I think it usefull as our minds see the migration folder but its not executed when starting the app 😁

mirdono: Indeed, the [migrations](https://github.com/mu-semtech/mu-migrations-service) service should be added to this app. Otherwise getting the migrations executed becomes trickier. :wink: Adding something like the following to `docker-compose.yml` should do the trick:

``` yaml
  migrations:
    image: semtech/mu-migrations-service:0.9.0
    links:
      - virtuoso:database
    volumes:
      - ./config/migrations:/data/migrations 
```

In the `docker-compose.dev.yml` an extra entry like the following should be added:

``` yaml
  migrations:
    restart: "no"
```

config/migrations/20240000000005-add-mock-users.sparql
```diff
@@ -0,0 +1,61 @@
+PREFIX besluit: <http://data.vlaanderen.be/ns/besluit#>
```
mirdono, line 1: **Hint:** You can use [mu-cli](https://github.com/mu-semtech/mu-cli) to generate migrations with a prefixed file name.

Usually we use timestamps up to and including seconds as prefix for migration files. If you use mu-cli and the migrations service the file creation is supported with a script:

``` shell
mu script migrations new sparql add-mock-users
```

Would create a blank file called: `20251017073706-add-mock-users.sparql** in the migrations folder.

config/migrations/20240000000005-add-mock-users.sparql
```diff
@@ -0,0 +1,61 @@
...
+ <http://data.lblod.info/id/identificatoren/a131fbc9-a78a-4c03-8dd3-bcd480c1810a>	<http://www.w3.org/2004/02/skos/core#notation>	"SharePoint identificator" .
+
+  }
+}
+WHERE {
+
+    VALUES (?g) {
+      (<http://mu.semte.ch/graphs/public>)
+      (<http://mu.semte.ch/graphs/organizations/7d7122bd82f165e091a36c830d041b59aa48e4be9ea0aa06dc4c67ef381b7505>)
+    }
```
mirdono, line 45: **Question:** Is is needed to duplicate all (administrative unit) data in both graphs?

I **think** only the `Person`/`OnlineAccount` data should be needed in the public graph, assuming that is where the mock-login component will look for candidates. The administrative unit data can than be limited to the organisation graph.

config/resources/user.lisp
```diff
@@ -0,0 +1,30 @@
...
+                         :inverse t
+                         :as "gebruiker"))
+  :on-path "accounts"
+)
+
+(define-resource bestuurseenheid () ;; Subclass of m8g:PublicOrganisation, which is a subclass of dct:Agent
+  :class (s-prefix "besluit:Bestuurseenheid")
+  :properties `((:naam :string ,(s-prefix "skos:prefLabel")))
+  :resource-base (s-url "http://data.lblod.info/id/bestuurseenheden/")
+  :on-path "bestuurseenheids"
```
mirdono, line 29: **Question:** Any reason not to use the correct plural "bestuurseenheden" here?

If changed the path in the dispatcher name will have to be updated as well. This might also ripple to the frontend, I recall ember-data requires something extra for pluralisation that is not just a +s.

config/resources/user.lisp
```diff
@@ -0,0 +1,30 @@
...
+  :resource-base (s-url "http://data.lblod.info/id/account/")
+  :properties `((:provider :via ,(s-prefix "foaf:accountServiceHomepage"))
+                (:vo-id :via ,(s-prefix "dct:identifier")))
+  :has-one `((gebruiker :via ,(s-prefix "foaf:account")
+                         :inverse t
+                         :as "gebruiker"))
+  :on-path "accounts"
+)
+
+(define-resource bestuurseenheid () ;; Subclass of m8g:PublicOrganisation, which is a subclass of dct:Agent
```
mirdono, line 25: **Question:** We do not need access to all data added for the administrative unit?

The migration adds more data for the administrative unit, also several identifiers. Since they are not added here they will not be accessible via resources. Is there a reason not to add here but do add them as data?

config/migrations/20240000000005-add-mock-users.sparql
```diff
@@ -0,0 +1,61 @@
...
+<http://data.lblod.info/id/bestuurseenheden/7d7122bd82f165e091a36c830d041b59aa48e4be9ea0aa06dc4c67ef381b7505>	<http://www.w3.org/2004/02/skos/core#prefLabel>	"Buggenhout" .
+<http://data.lblod.info/id/bestuurseenheden/7d7122bd82f165e091a36c830d041b59aa48e4be9ea0aa06dc4c67ef381b7505>	<http://mu.semte.ch/vocabularies/ext/kbonummer>	"0207445287" .
+<http://data.lblod.info/id/bestuurseenheden/7d7122bd82f165e091a36c830d041b59aa48e4be9ea0aa06dc4c67ef381b7505>	<http://data.vlaanderen.be/ns/besluit#werkingsgebied>	<http://data.lblod.info/id/werkingsgebieden/6c6dcdc5ea2a8eb3d951b9149e7979afa4c7f86143a7576ce9c480e33f26cd7d> .
+
+<http://data.lblod.info/id/bestuurseenheden/7d7122bd82f165e091a36c830d041b59aa48e4be9ea0aa06dc4c67ef381b7505>	<http://www.w3.org/1999/02/22-rdf-syntax-ns#type>	<http://data.vlaanderen.be/ns/besluit#Bestuurseenheid> .
+
+ <http://data.lblod.info/id/identificatoren/a131fbc9-a78a-4c03-8dd3-bcd480c1810a>	<https://data.vlaanderen.be/ns/generiek#gestructureerdeIdentificator>	<http://data.lblod.info/id/gestructureerdeIdentificatoren/2a7c78a9-8598-4741-b1eb-cae7084f1b9b> .
```
mirdono: **Issue:** The linked structured identifier is missing.

The linked structured identifier resource is not included in this migration. If this identifier data is not needed in this context I would remove it altogether. Otherwise, we should make sure to add complete and correct data.

config/resources/user.lisp
```diff
@@ -0,0 +1,30 @@
+(define-resource gebruiker ()
```
mirdono, line 1: **Question:** Should we try to avoid Dutch terminology?

Since DECIDe is a European project we might consider avoiding unnecessary Dutch terminology in our code base. But changing this will ripple through the several configuration files and to the frontend. It probably requires some discussion to see how much effort we put into that. Although it will only get more difficult to change at a later point if necessary.

@Rahien Do you have a (strong) opinion on this?

mirdono (changes requested):
I left one issue concerning the migration, as well as some minor questions and comments.

config/resources/user.lisp
niels-triply: Yes, I did this part together with @Rahien and we tried to add it to `config/custom-inflector-rules.js` but there was something else that made it not work (cannot quite recall at the moment). Iirc, we left it as-is because there would be changes needed that make `bestuurseenh{eids,eden}` obsolete in any case.

config/migrations/20240000000005-add-mock-users.sparql
niels-triply: If you do that, you get `{"errors":[{"title":"exactly one group should be linked"}]}` from `/mock/sessions`

config/resources/user.lisp
niels-triply: As I understand, the extra data is for the benefit of the mock-login service. I can see that it is not actually exposed through the dispatcher, but if I remove it (or replace with the simplified version below), we again get the `{"errors":[{"title":"exactly one group should be linked"}]}` error. @Rahien what was your interpretation again?

```sparql
#...
    ?account a foaf:OnlineAccount;
        mu:uuid ?uuidAccount;
        ext:sessionGroup ?admin;
        foaf:accountServiceHomepage <https://github.com/lblod/mock-login-service>;
        ext:sessionRole "Decide-gebruiker".
    ?admin a besluit:Bestuurseenheid;
        mu:uuid ?uuidAdmin.
#...
WHERE {
#...
    BIND(MD5(CONCAT(?volledigeNaam, "ADMIN")) as ?uuidAdmin)
    BIND(IRI(CONCAT("http://data.lblod.info/id/bestuurseenheid/", ?uuidAdmin)) AS ?admin)
}
```

config/migrations/20240000000005-add-mock-users.sparql
niels-triply: Okay, I have removed that triple. I do not know exactly how the rest of the triples are used, but as discussed elsewhere in this PR, I cannot quite remove them without causing errors, so I think that should be addressed in a new ticket

config/migrations/20240000000005-add-mock-users.sparql
mirdono: I did some testing for this, along with some digging in the mock-login service. My initial remark is indeed not correct. We do need some triples for the administrative unit resource in the public graph, namely its resource type and UUID triple. The mock-login service uses that data to [determine the correct organisation-specific graph](https://github.com/lblod/mock-login-service/blob/master/login_service/sparql_queries.rb#L78). The other triples for the administrative unit resource are not necessary in the public graph (I think we do not even need them in the organisation graph, but let's keep them there to have something resembling actual data.)

To make it very concrete, the triples we need in the public graph for the administrative unit resource are:

``` sparql
<http://data.lblod.info/id/bestuurseenheden/7d7122bd82f165e091a36c830d041b59aa48e4be9ea0aa06dc4c67ef381b7505> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://data.vlaanderen.be/ns/besluit#Bestuurseenheid> .
<http://data.lblod.info/id/bestuurseenheden/7d7122bd82f165e091a36c830d041b59aa48e4be9ea0aa06dc4c67ef381b7505> <http://mu.semte.ch/vocabularies/core/uuid> "7d7122bd82f165e091a36c830d041b59aa48e4be9ea0aa06dc4c67ef381b7505" .
```

mirdono (commented):
Some minor comments, imo only the comment about using tagged version for the mock-login service is blocking. The others are non-blocking. Personally, I am inclined to avoid inserting unneeded data in the public graph as this data tends to stick around (who remembers to clean up old test data...) and can cause confusion over time.

docker-compose.dev.yml
```diff
@@ -14,4 +14,8 @@ services:
...
+    restart: "no"
+  mocklogin:
+    image: lblod/mock-login-service
```
mirdono: **Remark:** Explicitly specify a version tag.

I would explicitly tag the used version, most recent release is `v0.7.0`. That way everybody is using the same version of the service in their development environment. This avoids that people end up using different `latest` variants over time. (Due to the limited amount of changes for this service this is less likely in this specific case. But it is common practice in semantic.works stacks to use tagged release unless that is not possible for some reason.)

config/migrations/20240000000005-add-mock-users.sparql
```diff
@@ -0,0 +1,60 @@
...
+        mu:uuid ?uuidAccount;
+        ext:sessionGroup <http://data.lblod.info/id/bestuurseenheden/7d7122bd82f165e091a36c830d041b59aa48e4be9ea0aa06dc4c67ef381b7505>;
+        foaf:accountServiceHomepage <https://github.com/lblod/mock-login-service>;
+        ext:sessionRole "Decide-gebruiker".
+
+<http://data.lblod.info/id/bestuurseenheden/7d7122bd82f165e091a36c830d041b59aa48e4be9ea0aa06dc4c67ef381b7505>	<http://www.w3.org/ns/org#classification>	<http://data.vlaanderen.be/id/concept/BestuurseenheidClassificatieCode/5ab0e9b8a3b2ca7c5e000001> .
+
+<http://data.lblod.info/id/bestuurseenheden/7d7122bd82f165e091a36c830d041b59aa48e4be9ea0aa06dc4c67ef381b7505>	<http://data.vlaanderen.be/ns/besluit#classificatie>	<http://data.vlaanderen.be/id/concept/BestuurseenheidClassificatieCode/5ab0e9b8a3b2ca7c5e000001> .
+
+<http://data.lblod.info/id/bestuurseenheden/7d7122bd82f165e091a36c830d041b59aa48e4be9ea0aa06dc4c67ef381b7505>	<http://www.w3.org/ns/org#hasPrimarySite>	<http://data.lblod.info/id/vestigingen/f34a68fdd29a73b4bb1e394bfedb670a> .
```
mirdono: **Remark:** The linked resource is not further defined.

Since we do not include the data for the actual resource, I do not think it is needed to link it here.

config/migrations/20240000000005-add-mock-users.sparql
```diff
@@ -0,0 +1,60 @@
...
+
+<http://data.lblod.info/id/bestuurseenheden/7d7122bd82f165e091a36c830d041b59aa48e4be9ea0aa06dc4c67ef381b7505>	<http://www.w3.org/ns/regorg#legalName>	"Buggenhout" .
+<http://data.lblod.info/id/bestuurseenheden/7d7122bd82f165e091a36c830d041b59aa48e4be9ea0aa06dc4c67ef381b7505>	<http://purl.org/dc/terms/identifier>	"OVO002038" .
+
+<http://data.lblod.info/id/bestuurseenheden/7d7122bd82f165e091a36c830d041b59aa48e4be9ea0aa06dc4c67ef381b7505>	<http://mu.semte.ch/vocabularies/core/uuid>	"7d7122bd82f165e091a36c830d041b59aa48e4be9ea0aa06dc4c67ef381b7505" .
+<http://data.lblod.info/id/bestuurseenheden/7d7122bd82f165e091a36c830d041b59aa48e4be9ea0aa06dc4c67ef381b7505>	<http://www.w3.org/2004/02/skos/core#prefLabel>	"Buggenhout" .
+<http://data.lblod.info/id/bestuurseenheden/7d7122bd82f165e091a36c830d041b59aa48e4be9ea0aa06dc4c67ef381b7505>	<http://mu.semte.ch/vocabularies/ext/kbonummer>	"0207445287" .
+<http://data.lblod.info/id/bestuurseenheden/7d7122bd82f165e091a36c830d041b59aa48e4be9ea0aa06dc4c67ef381b7505>	<http://data.vlaanderen.be/ns/besluit#werkingsgebied>	<http://data.lblod.info/id/werkingsgebieden/6c6dcdc5ea2a8eb3d951b9149e7979afa4c7f86143a7576ce9c480e33f26cd7d> .
```
mirdono: **Remark:** The linked resource is not further defined.

Since we do not include the data for the actual resource, I do not think it is needed to link it here.

config/resources/user.lisp
Rahien: i agree, we should translate this to english. However, the mock-login service may require the dutch version, since it uses the ember models. @niels-triply can you see what happens if we translate?

config/resources/user.lisp
Rahien: Indeed the mock-login service needs this data, but (for now?) the application does nothing with the information on the bestuurseenheid (group) itself. I would keep this for now and see how we will use groups in the app later on

mirdono: Thanks for the responses, I will mark this resolved as my initial question has been answered.

config/migrations/20240000000005-add-mock-users.sparql
niels-triply: Ah, you're right. That is essentially what I did for https://github.com/lblod/app-decide/pull/11#discussion_r2439896415 but I had apparently forgotten the triple for the user, because now it does work. 

The other graph can indeed be left out then. My intuition is then to do so, for the reason you also gave:

> I am inclined to avoid inserting unneeded data in the public graph as this data tends to stick around

config/resources/user.lisp
niels-triply: Yeah, you would like to keep it? I would remove it, cf https://github.com/lblod/app-decide/pull/11#discussion_r2447737837

mirdono: The resources configuration as written here can stay as is, based on your responses the frontend does not need access to the additional administrative unit data added by the migration. If that changes in the future we can always update the configuration accordingly.

In retrospect, I should have phrased my initial question better. It should have been _"Does the **frontend** not need access to all (or more) data added for the administrative unit?"_ The usage of _"we"_ in my initial phrasing is too ambiguous for this discussion. Your responses answered that, rephrased, question for me with a "no, the frontend does not need that." Hence I consider this thread resolved.

config/migrations/20240000000005-add-mock-users.sparql
mirdono: To avoid confusion, I would propose inserting the data as follows:
- Public graph: everything on `foaf:Person` and `foaf:OnlineAccount` resources, but for the `besluit:Bestuurseenheid` resource only the UUID and type triples. Note, I am not sure we actually need the first two resources here. A quick look at the queries performed by the mock-login service along with a quick local test seems to indicate that mock-login works without those resources in the public graph.
- Organisation graph: everything on the `foaf:Person`, `foaf:OnlineAccount`, and `besluit:Bestuurseenheid` resources. (With the exception of the triples linking to undefined resources mentioned in other comments.)

Note that we cannot just drop the organisation graph as the mock-login service will search in there for `foaf:Person` and `foaf:OnlineAccount`, as can be seen in the code I linked in my previous reply.

config/resources/user.lisp
niels-triply: Translating into English doesn't immediately work for reasons I do not fully comprehend, so I'm inclined to do that as a separate task.

config/migrations/20240000000005-add-mock-users.sparql
niels-triply: Ah yes, forgot that the session gets invalidated without the organisation graph. I've put the data minus the previously mentioned triples (and made it a bit more readable). Thanks for your help.

niels-triply:
@mirdono Can I merge this after [a6eacb6](https://github.com/lblod/app-decide/pull/11/commits/a6eacb625fb8d7895ea1783ad4cc0da84c650492) and [662ec28](https://github.com/lblod/app-decide/pull/11/commits/662ec288e5939c58ec3d7fb29f67a1768dfdd9e4)?

I assume so since you indicated only the latter was blocking in the first place, but I have noticed that session gets invalidated on Firefox. Separate ticket for that? (Should be an issue on the front end in any case)

mirdono (approved):
Looks good to me. I agree that we can tackle to translation etc. as separate tickets.

PR #13, opened 2025-10-17
Feature/harvester oparl

brechtvdv:


docker-compose.dev.yml
```diff
@@ -19,7 +19,7 @@ services:
       - debug
   virtuoso:
     ports:
-      - "8890:8890"
+      - "8891:8890"
```
Rahien: can you undo this change and put it in you docker.compose.override.yml?

docker-compose.yml
```diff
@@ -85,3 +85,64 @@ services:
...
+    environment:
+      MU_APPLICATION_GRAPH: "http://mu.semte.ch/graphs/harvesting"
+    labels:
+      - "logging=true"
+    restart: always
+    logging: *default-logging
+  harvest_diff:
+    image: lblod/harvesting-diff-service:0.3.0
+    environment:
+      # SLEEP_BETWEEN_JOB_MS: 100 --> replaced by a semaphore (max nbcpu conccurent tasks)
```
Rahien: do we need this comment?

docker-compose.yml
```diff
@@ -85,3 +85,64 @@ services:
       CREDENTIAL_TYPE: DecideCredential
       NO_DID_PREFIX: true
       SINGLE_CREDENTIAL_RESPONSE: true
+    restart: always
+    labels:
+      - "logging=true"
+  deltanotifier:
+    image: semtech/mu-delta-notifier:0.0.4
```
Rahien: should be 0.4.0

docker-compose.dev.yml
Rahien: you accidentally resolved this but didn't actually put the port back to 8890, there are people that use plugins in different editors that count on this port and changing it in each of their environments is a hassle

Rahien: if you want this to be different, docker-compose.override.yml is the way to go

docker-compose.yml
```diff
@@ -29,6 +29,8 @@ services:
     restart: always
     logging: *default-logging
   virtuoso:
+    ports:
```
Rahien: this must be removed. we should never ever expose the virutoso in production

Rahien (changes requested):

Rahien (approved):

PR #15, opened 2025-10-30
Fix syntax in dispatcher file.

niels-triply:
The frontend worked, but only as long as it only does GET requests at root level, ie. `/catalog?filter...` works but `/catalog/<uuid>` did not. The cause was the lack of postfix slashes in the dispatcher config.


mirdono (approved):
Thanks for fixing this. My apologies for the extra work this caused, in retrospect I should have noticed this in #7.

PR #16, opened 2025-11-04
basic search tryout with sample decisions

Rahien:
WIP. don't base yourself off this or do anything with it atm


config/search/config.json
```diff
@@ -0,0 +1,142 @@
...
+    "number_of_replicas": 0,
+    "analysis": {
+      "analyzer": {
+        "dutch_analyzer": {
+          "tokenizer": "standard",
+          "filter": ["lowercase", "asciifolding", "dutch_stemmer", "dutch_stop"]
+        },
+        "german_analyzer": {
+          "tokenizer": "standard",
+          "filter": ["lowercase", "asciifolding", "german_stemmer"]
```
MartijnBogaert: @Rahien Is there a reason not to have a German stop token filter?

Rahien: yes! complete oversight on my part while playing with the ES. 
To be fair, I don't consider this setup to be the best possible one at all (for instance, there are only two types atm and we will have many more). There is a separate ticket to properly fill (and configure, i will make that explicit) mu-search for our case

MartijnBogaert: Makes sense!

Rahien:
this test was not meant to be merged. closing

PR #18, opened 2025-11-12
Consume Lokaal Beslist dump

MartijnBogaert:
This consumer setup performs an initial sync of the Lokaal Beslist harverst 1 (as it holds decision data coming from Ghent bestuursorganen).
The delta sync step is intentionally disabled as it will be tackled later on. The goal for now is simply to have data available.

### DIFF
--- a/config/virtuoso/virtuoso.ini
+++ b/config/virtuoso/virtuoso.ini
@@ -37,7 +37,7 @@ TempStorage			= TempDatabase
 [TempDatabase]
 DatabaseFile			= /usr/local/virtuoso-opensource/var/lib/virtuoso/db/virtuoso-temp.db
 TransactionFile			= /usr/local/virtuoso-opensource/var/lib/virtuoso/db/virtuoso-temp.trx
-MaxCheckpointRemap		= 2000
+MaxCheckpointRemap		= 4000
 Striping			= 0
 
 
@@ -77,7 +77,7 @@ MaxMemPoolSize                  = 200000000
 PrefixResultNames               = 0
 MacSpotlight                    = 0
 IndexTreeMaps                   = 64
-MaxQueryMem 		 	= 2G		; memory allocated to query processor
+MaxQueryMem 		 	= 4G		; memory allocated to query processor
 VectorSize 		 	= 1000		; initial parallel query vector (array of query operations) size
 MaxVectorSize 		 	= 1000000	; query vector size threshold.
 AdjustVectorSize 	 	= 0
@@ -113,8 +113,8 @@ AsyncQueueMaxThreads 	 	= 10
 ;; Note the default settings will take very little memory
 ;; but will not result in very good performance
 ;;
-NumberOfBuffers          = 10000
-MaxDirtyBuffers          = 6000
+NumberOfBuffers          = 680000
+MaxDirtyBuffers          = 500000
 
 
 [HTTPServer]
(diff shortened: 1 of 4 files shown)


MartijnBogaert:
@nbittich I tagged you as well in the hopes for you to have a quick look at the `virtuoso.ini` file. It is the same one you set up a couple of weeks back when you were setting up a consumer config as well. Does it still hold in this situation?

config/virtuoso/virtuoso.ini
```diff
@@ -236,10 +236,10 @@ DefaultHost			= localhost:8890
...
 ;DefaultGraph      		= http://localhost:8890/dataspace
 ;ImmutableGraphs    		= http://localhost:8890/dataspace
-ResultSetMaxRows           	= 10000
-MaxConstructTriples      	  = 10000
-MaxQueryCostEstimationTime 	= 400	; in seconds
-MaxQueryExecutionTime      	= 60	; in seconds
+ResultSetMaxRows           	= 10000000
+MaxConstructTriples      	  = 10000000
+MaxQueryCostEstimationTime 	= 4000	; in seconds
+MaxQueryExecutionTime      	= 1200	; in seconds
```
Rahien: we could also disable these: https://docs.openlinksw.com/virtuoso/ch-server/
resultsetmaxrows can be set to 0 for no limit and if you don't specify max execution times the default is no limit

Rahien (approved):

nbittich:
ah yes idd, it came from lokaalbeslist prod and has  been tweaked to handle the consumer loads. I think it might be relevant for decide as well, iirc the consumer won't run successfully without it

config/virtuoso/virtuoso.ini
MartijnBogaert: I disabled the `MaxQueryCostEstimationTime` and `MaxQueryExecutionTime` rules, as the provided documentation indeed points out their defaults are _no limit_.
However, I couldn't find anything on `ResultSetMaxRows` having any default, so I'm a bit afraid assigning it to 0.

Rahien: yes they managed to hide it quite successfully in their docs. it's here (bold added myself for clarity):
> MaxMemInUse = 0 .  Maximum amount of memory that is allowed for temporary storing parts of a SPARQL query result. Default is zero for no limit. **ResultSetMaxRows is maximum length of a SPARQL query result. Default is zero for no limit**. These two options may be useful for protecting public web service endpoints against DOS attacks, but at the same time it may cause returning incomplete results without reporting errors. When used, it is strongly advised to set the value orders of magnitude larger than the expected size of longest reply. As a rule of thumb, timeout should happen before this limit has reached. Values less than 1000000 bytes are impractical in all cases.

MartijnBogaert: Good catch! I updated it.

MartijnBogaert:
@Rahien I updated three things since your last review:
- I implemented your proposed changes in `virtuoso.ini`.
- The initial sync doesn't start automatically anymore, but should be enabled in override (explained in readme), to prevent users unknowingly downloading gigabytes of data upon upping the stack.
- I noticed the initial sync needed more time than expected on the server, making me realize I mistakenly used `BATCH_SIZE` instead of `DCR_BATCH_SIZE` in docker compose 🙈, that's been fixed now.

If you agree with the changes, I think I can merge.

Rahien (approved):

PR #27, opened 2025-12-05
feat(authz): add geographical and annotation to public graph

mirdono:
Extend the definition for the public graph with type specifications for geographical and annotation data resources. This makes such kinds of resources publicly readable for later use.

## Related tickets
- LBRON-918


mirdono:
**Question:** Should this data be placed in the `public` graph?

For the moment I opted to assume that these resources would be in the `public` graph. But I am not certain whether this is correct.
On the DEV environment, the ELI(-DL) data is currently located in the `harvesting` graph, and some in the `oparl-temp` and `oslo-temp`.
Whereas the existing sparql-parser config puts ELI(-DL) in the `public` graph.
In it not clear which of these graphs is currently meant to be used for this kind of data? The `public` graphs seems the more logical option as the `harvesting` graph is used by the oparl-to-eli pipeline.
Based on this I, for now, assumed that the location data will also be put in the same graph. The gitbook page does mention a discussion about a seperate "location database" but it is not clear whether any decisions were already made on this. Similarly, I assumed annotations will be put in the same graph as the resources they are linked to.

brechtvdv:
Oparl data should indeed land in the `public` graph. Currently this is the `harvesting` graph, because I will probably need another service to do some diffing for the latest updates

Rahien (approved):

mirdono:
> Oparl data should indeed land in the `public` graph. Currently this is the `harvesting` graph, because I will probably need another service to do some diffing for the latest updates

In that case I will make the ELI(-DL) that DEV currently has in its `harvesting` graph also publicly readable there. That should simplify developing the PoC frontend showing this data.

config/authorization/config.lisp
```diff
@@ -23,29 +23,32 @@
 (define-prefixes
   :besluit "http://data.vlaanderen.be/ns/besluit#"
   :cms "http://mu.semte.ch/vocabulary/cms/"
+  :cogs "http://vocab.deri.ie/cogs#"
+  :core "http://open-services.net/ns/core#"
   :dcat "http://www.w3.org/ns/dcat#"
   :dct "http://purl.org/dc/terms/"
+  :defend "https://d3fend.mitre.org/ontologies/d3fend#"
   :eli "http://data.europa.eu/eli/ontology#"
   :eli-dl "http://data.europa.eu/eli/eli-draft-legislation-ontology#"
```
mirdono, line 29: **Issue:** The `oparl-to-eli-service` uses a different expansion for the `eli-dl` prefix.

The `oparl-to-eli-service` uses `http://data.europa.eu/eli/ontology/dl#Activity` as expansion, see

[here](https://github.com/lblod/oparl-to-eli-service/blob/main/constants.ts#L64). Consequently the DEV data also uses this one, making the configuration incompatible. Should these be the same or are they different data models?

brechtvdv: It should indeed be: <http://data.europa.eu/eli/eli-draft-legislation-ontology#>
I'll fix in oparl-to-eli-service

mirdono: Perfect, thanks for confirming.

brechtvdv: Fixed: https://github.com/lblod/oparl-to-eli-service/pull/8

PR #28, opened 2025-12-10
feat: define DECIDe authorisation configuration as ODRL policy

mirdono:
Updates the example ODRL authorisation policy to one matching the actual authorisation configuration of the DECIDe app. The previous policy was fictitious one loosely based on parts of the LMB app with some additions to illustrate the possibilities with policies in ODRL.[^1]

## Using the generated file as policy
Updating a generated configuration is currently rather cumbersome. More specifically, it involves four steps:

1. Update the `config/odrl-parser/config.ttl` file
2. Convert the modified file to an N-triples file that `odrl-parser` can parse using the project script: `mu script project-scripts ttl-to-ntriples -r`
3. Load the updated policy into triplestore: `docker compose exec odrl-parser curl http://localhost/load-policy`
4.  Generate policy file: `docker compose exec odrl-parser curl http://localhost/generate-config`

Because this process is rather labour intensive, and potentially fragile, I opted to not yet default to using the generated configuration as authorisation policy. This is also because I expect DECIDe's authorisation policy is likely to change somewhat frequently as the app is further developed. I propose to actively start using the ODRL policy once the ODRL policy functionality is integrated into `sparql-parser` itself, which should provide a better user experience.

To still allow experimenting with generated authorisation configurations I did add some functionality to the app's `sparql-parser` setup. The configuration in `config/authorization/config.lisp` no longer contains the actual policy but loads one of two files based on a new environment variable `USE_ODRL_POLICY`. If this environment variable is set to `"true"` the generated policy in the `config/odrl-parse/decideAuthorizationPolicy.lisp` file will be loaded. Otherwise, the policy in `config/authorization/config.lisp` will be loaded, which contains the app's policy as it was manually defined so far.

## Differences between manual and generated configuration
There are some differences between the manually written and generated configuration:

- The order of elements is different.
- The defined set of prefixes differs because of two reasons:
  + Not all prefixes defined in the manual configuration are actually used in it. For example, `` and ``. I did not port the unused prefixes to the ODRL policy.
  + Internally the `odrl-parser` (and its dependencies) load some default prefixes, e.g. `odrl:` and `rdf:`. Although they are not actually used in the generated configuration they are exported during the generation.

## Related tickets
- LBRON-880

[^1]: For those interested, the example policy is still available in the `odrl-parser-service` [repository](https://github.com/lblod/odrl-parser-service/tree/master/examples).


docker-compose.yml
```diff
@@ -21,8 +21,11 @@ services:
     logging: *default-logging
   database:
     image: semtech/sparql-parser:0.0.14
+    environment:
+      USE_ODRL_POLICY: "false"
     volumes:
       - ./config/authorization:/config
+      - ./config/odrl-parser:/odrl
```
mirdono, line 28: **Comment:** This additional volume allows to load the generated configuration.

An alternative is to have `odrl-parser` to write its generated file to one of the existing volumes. But that would add extra complexity to that service to properly handle multiple, possibly non-existing, locations. Furthermore, it can potentially cause problems when `odrl-parser` (incorrectly) overwrites necessary files. Adding a volume to the `sparql-parser` service seemed the easier less breakable solution.

docker-compose.yml
```diff
@@ -63,7 +66,7 @@ services:
   dcat:
     image: lblod/dcat-service:0.0.1
   odrl-parser:
-    image: lblod/odrl-parser-service:0.0.3
+    image: lblod/odrl-parser-service:0.0.5
```
mirdono, line 69: **Comment:** Bumped the `odrl-parser` to most recent version.

While working on updating the ODRL policy and regenerating the corresponding configuration I came across some issues with the `odrl-parser` service. I fixed/improved this issues in the service, see its [changelog](https://github.com/lblod/odrl-parser-service/blob/master/CHANGELOG.md) for details.

Rahien (approved):
generated config is indeed the same and no issues running mu-auth

PR #29, opened 2025-12-11
Introduce annotation jobs data model

MartijnBogaert:
LBRON-725

Datamodel described here: https://app.gitbook.com/o/-MP9Yduzf5xu7wIebqPG/s/grRYjcouX0uZY2P6PYBl/data-analysis-research/datamodel-annotation-jobs

Everything necessary has been added to resource, mu-auth and dispatcher configs.


config/resources/annotation.lisp
```diff
@@ -0,0 +1,11 @@
+(in-package :mu-cl-resources)
+
+(define-resource annotation ()
+  :class (s-prefix "oa:Annotation")
+  :properties `((:motivated-by :url ,(s-prefix "oa:motivatedBy"))
+                (:target :url ,(s-prefix "oa:hasTarget"))
+                (:body :url ,(s-prefix "oa:hasBody"))
```
Rahien, line 7: this means our frontend will not be able to do much with this though right? It doesn't know the type etc... How will we render these things?

Rahien: still this was not the point of this PR, but it is a question we should get answered when starting on the human review ui

Rahien (approved):

config/resources/annotation.lisp
MartijnBogaert: When we build the frontend purely for annotation jobs, we can assume the target to be an `eli:Expression` or `eli:Work` and the body to be a `skos:Concept`. However, the resource config keeps things general since annotations can of course also appear in other contexts and we want to account for those.

PR #30, opened 2025-12-15
Add frontend-harvester-self-service (Feature/add-job-button)

brechtvdv:
Features:
- replaces LBLOD dashboard with frontend-harvester-self-service (supports adding and deleting jobs)
- Adds login service with test user
- Adds singleton job service for compatibility with lblod harvester


config/migrations/add-test-user/20251211000000-add-test-user.sparql
MartijnBogaert, line 1: Probably a good idea to rename this to something like "20251211000000-add-test-user.sparql" as it's a bit of common practice for migrations naming.

MartijnBogaert (changes requested):
Everything works! Also the deleting 😉
I just left a comment regarding the file name of the migration.

MartijnBogaert:
I see you'll have to resolve conflicts, one of them is the auth config: the `config.lisp` should no longer hold specific config rules, those should be moved to `decide.lisp`.

config/migrations/add-test-user/20251211000000-add-test-user.sparql
brechtvdv: Fixed

MartijnBogaert (approved):

PR #33, opened 2026-01-08
Add scheduled job routes and controller

brechtvdv:
- Adds scheduled-job-controller-service
- Adds delta config for scheduled job controller
- Adds support for scheduled jobs and tasks in mu-resources and mu auth


MartijnBogaert (approved):
Looks good!

PR #34, opened 2026-01-12
Add migration for SDG codelist

brechtvdv:


Rahien:
you're adding the full codelist as published by the un here. That's likely the proper way of doing this, but it means that our matching service will need to have an understanding of skos:broader etc so it can 'cut' to a certain level as it is unlikely that we want to match the lower level terms to decisions (lower success rate, start small, ...)

another way around this is to create a simpler code list with only the top level concepts.

brechtvdv:
It's up to the Task description and service to support deeper levels of the codelist. I don't see an issue in keeping all levels of the SDGs?

PR #35, opened 2026-01-12
Add missing prefixes

madnificent:
Adds a few missing prefixes.  These are the defaults from prefix.cc when found, or what the login service prescribes.

### DIFF
--- a/config/resources/repository.lisp
+++ b/config/resources/repository.lisp
@@ -24,3 +24,6 @@
 (add-prefix "dgftOauth" "http://kanselarij.vo.data.gift/vocabularies/oauth-2.0-session/")
 (add-prefix "geosparql" "http://www.opengis.net/ont/geosparql#")
 (add-prefix "locn" "http://www.w3.org/ns/locn#")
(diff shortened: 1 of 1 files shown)


mirdono:
Could you elaborate what issue this solves/improves or what triggered this PR? At least for me, the description does not explain why these changes are necessary.

madnificent:
These prefixes are used in the domain files but they're not available.  Using this part of the API will throw an error.  Either the API space is more broadly defined than intended, or this part of the API is not functional.  I don't know the application sufficiently well to know which it is.

This was triggered by efforts on DiSHACLed through which we annotate mu-cl-resources automatically with the shapes it uses.  This could aid the discovery of related services and it could support charting data flows.  These errors pop up when running `mu project doc` which also gives you an SVG of the json-api.

The change is not urgent and we don't know of hacks possible through this configuration, though it's probably best to have a fully defined configuration.  The configuration has to be complete if we intend to scan automatically.

Rahien (approved):
merging! i could run mu project doc successfully even with missing prefixes though

PR #12, opened 2025-10-16
Add DCAT mock data.

niels-triply:
Used in https://github.com/lblod/frontend-decide/pull/2

PR #37, opened 2026-01-19
Add real DCAT datasets, distributions and services

niels-triply:
References:

- https://binnenland.atlassian.net/browse/LBRON-339
- https://binnenland.atlassian.net/browse/LBRON-879

These tickets both refer to DCAT descriptions, so they are taken together in this PR. As discussed with @Rahien , I have additionally included the other datasets collected on GitBook.

In absence of https://binnenland.atlassian.net/browse/LBRON-875, I have passed these through OSLO Application Profile's SHACL validation, which can be found here: https://data.vlaanderen.be/doc/applicatieprofiel/metadata-dcat/erkendestandaard/2022-04-21/

However, this leads to a lot of violations, some of which can be ignored. In particular, the reasoning is as follows:

- It demands that `dct:format` is drawn from <https://publications.europa.eu/resource/authority/file-type>, but the canonical URLs use the HTTP scheme, not HTTPS.
- OSLO would like us to draw `dct:conformsTo` from <https://data.vlaanderen.be/id/conceptscheme/dataserviceprotocol> but if we did that, we wouldn't be able to describe OParl; <http://publications.europa.eu/resource/authority/file-type> seems more applicable to an European context anyway.
- It complains that we have not set a `dcat:downloadURL` and `dc:identifier`, which is not required.
- I am used to link to external web pages (that shouldn't be interpreted as RDF resources) as literal with data type `xsd:anyURI`, instead of as URI resource. These shapes reject that, which suggests that this is not a convention that we follow within ABB. Is that the case?

### DIFF
--- a/config/migrations/add-decide-datasets/20260108152100-overview.ttl
+++ b/config/migrations/add-decide-datasets/20260108152100-overview.ttl
@@ -30,27 +30,24 @@ agent:7365f73a-b57c-4999-aa8b-61464d9d3c6f a foaf:Agent;
 .
 
 agent:658900ce-63cf-4d45-8fd5-a67ed8deab2c a foaf:Agent;
-    vcard:hasName "Municipality of Freiburg"@en
-.
-
-agent:c3c2a7bf-af3b-43b8-99f5-c3463ec9408a a foaf:Agent;
-    vcard:hasName "Vlaamse Overheid"@nl, "Flemish government"@en
+    vcard:hasName "Municipality of Freiburg"@en;
+    vcard:hasEmail "rsk-ratsbuero@freiburg.de"
 .
 
 agent:f45c1220-b250-4b93-8a8a-3e7e4a12a467 a foaf:Agent;
-    vcard:hasName "ABB Open Proces Huis"@nl
-.
-
-agent:436c5bce-8380-401e-8027-f34ef4acd97e a foaf:Agent;
-    vcard:hasName "ABB Lokaal Beslist"@nl
+    vcard:hasName "Agentschap Binnenlands Bestuur"@nl;
+    vcard:hasEmail "binnenland@vlaanderen.be"
 .
 
 agent:7cb46783-0ca2-41a2-b28c-248b99e752e4 a foaf:Agent;
-    vcard:hasName "ABB DECIDe"@nl
+    vcard:hasName "Agentschap Binnenlands Bestuur (DECIDe)"@nl;
+    vcard:hasEmail "binnenland@vlaanderen.be";
+    rdfs:comment "This is for datasets published within the DECIDe project; the actual entity and contact data may be subject to change."@en
 .
 
-agent:bcdaccd2-5651-4c1d-8021-415a43ee4cfb a foaf:Agent;
-    vcard:hasName "ABB DECIDe AI Team"@nl
+agent:436c5bce-8380-401e-8027-f34ef4acd97e a foaf:Agent;
+    vcard:hasName "Lokaal Beslist"@nl;
+    vcard:hasEmail "lokaalbeslist@vlaanderen.be"
 .
 
 agent:945e9df0-c9e6-4f5c-a249-34075169e68d a foaf:Agent;
@@ -218,8 +215,8 @@ dataset:dd91b174-e9bb-4d33-8732-ed064a828b60 a dcat:Dataset;
     mu:uuid "dd91b174-e9bb-4d33-8732-ed064a828b60";
     dct:title "Ghent Municipal Decisions Roadblocks Classification"@en;
     dct:description "A comprehensive dataset of Dutch municipal decision documents from Ghent, Belgium, specifically processed and classified for traffic measures, roadblocks, and infrastructure projects that affect urban mobility."@en;
-    dct:publisher agent:bcdaccd2-5651-4c1d-8021-415a43ee4cfb;
-    dcat:contactPoint agent:bcdaccd2-5651-4c1d-8021-415a43ee4cfb;
+    dct:publisher agent:7cb46783-0ca2-41a2-b28c-248b99e752e4;
+    dcat:contactPoint agent:7cb46783-0ca2-41a2-b28c-248b99e752e4;
     dcat:theme <http://vocab.belgif.be/auth/datatheme/REGI>;
     dct:accessRights <http://publications.europa.eu/resource/authority/access-right/PUBLIC>;
     dcat:distribution distribution:f27a58aa-5cbf-4612-b3d0-a4d49100153f
@@ -237,8 +234,8 @@ dataset:2e1d52d9-2509-4b60-ac5a-0307897d64ef a dcat:Dataset;
     mu:uuid "2e1d52d9-2509-4b60-ac5a-0307897d64ef";
     dct:title "llama3 ABB Instruct"@en;
     dct:description "A comprehensive dataset to fine tune a llama3 model for translation, classification and keyword extraction from decision documents from Ghent"@en;
-    dct:publisher agent:bcdaccd2-5651-4c1d-8021-415a43ee4cfb;
-    dcat:contactPoint agent:bcdaccd2-5651-4c1d-8021-415a43ee4cfb;
+    dct:publisher agent:7cb46783-0ca2-41a2-b28c-248b99e752e4;
+    dcat:contactPoint agent:7cb46783-0ca2-41a2-b28c-248b99e752e4;
     dcat:theme <http://vocab.belgif.be/auth/datatheme/GOVE>;
     dct:accessRights <http://publications.europa.eu/resource/authority/access-right/PUBLIC>;
     dcat:distribution distribution:4ed47dd1-aa35-4f94-b751-c0984349908c
@@ -256,8 +253,8 @@ dataset:6aaa542d-410c-4150-bcfd-e5526eab87f4 a dcat:Dataset;
     mu:uuid "6aaa542d-410c-4150-bcfd-e5526eab87f4";
     dct:title "ABB DECIDe SDGs Antwerp"@en;
     dct:description "A comprehensive dataset to classify decision documents from Antwerp into Sustainable Development Goals"@en;
-    dct:publisher agent:bcdaccd2-5651-4c1d-8021-415a43ee4cfb;
-    dcat:contactPoint agent:bcdaccd2-5651-4c1d-8021-415a43ee4cfb;
+    dct:publisher agent:7cb46783-0ca2-41a2-b28c-248b99e752e4;
+    dcat:contactPoint agent:7cb46783-0ca2-41a2-b28c-248b99e752e4;
     dcat:theme <http://vocab.belgif.be/auth/datatheme/ENVI>;
     dct:accessRights <http://publications.europa.eu/resource/authority/access-right/PUBLIC>;
     dcat:distribution distribution:6d5607a1-e1b3-4259-b186-760d8fef7da3
@@ -275,8 +272,8 @@ dataset:3d79b143-4009-451a-8b80-423da944d9a8 a dcat:Dataset;
     mu:uuid "3d79b143-4009-451a-8b80-423da944d9a8";
     dct:title "ABB DECIDe title extraction instruct"@en;
     dct:description "A comprehensive dataset to identify and extract the title from plain text decision documents from Ghent with the help of llama3"@en;
-    dct:publisher agent:bcdaccd2-5651-4c1d-8021-415a43ee4cfb;
-    dcat:contactPoint agent:bcdaccd2-5651-4c1d-8021-415a43ee4cfb;
+    dct:publisher agent:7cb46783-0ca2-41a2-b28c-248b99e752e4;
+    dcat:contactPoint agent:7cb46783-0ca2-41a2-b28c-248b99e752e4;
     dcat:theme <http://vocab.belgif.be/auth/datatheme/GOVE>;
     dct:accessRights <http://publications.europa.eu/resource/authority/access-right/PUBLIC>;
     dcat:distribution distribution:b8b63c96-f5ce-496a-937e-7406f6aa1b91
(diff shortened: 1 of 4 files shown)


mirdono (commented):
Left some questions and nitpicks in inline comments.


Concerning the use of `xsd:anyURI`, I am not aware of guidelines on this and have not noticed it being used consistently, but this is anecdotal based on my personal experience.

One more general remark:
**Remark:** Should we provide English language strings where there are currently on Dutch ones.

Several titles, descriptions, and comments are currently only Dutch language strings. Considering DECIDe is a European project, should we not provide English ones next to those?

config/migrations/add-decide-datasets/20260108152100-overview.ttl
```diff
@@ -0,0 +1,520 @@
...
+    dcat:accessURL "https://openproceshuis.vlaanderen.be/sparql"^^xsd:anyURI;
+    dcat:accessService service:7333d62b-bd34-4063-ad34-28ec22a098b5
+.
+service:7333d62b-bd34-4063-ad34-28ec22a098b5 a dcat:DataService ;
+    mu:uuid "7333d62b-bd34-4063-ad34-28ec22a098b5";
+    dct:title "Open Proces Huis procesdata"@nl;
+    dct:conformsTo <https://www.w3.org/TR/sparql11-protocol/> ;
+    dcat:endpointDescription "Proces-, diagram- en invatarisdata van Open Proces Huis"@nl;
+    dcat:endpointURL "https://openproceshuis.vlaanderen.be/sparql"^^xsd:anyURI ;
+    dct:accessRights <http://publications.europa.eu/resource/authority/access-right/PUBLIC>;
```
mirdono: **Question:** Is this actually public data?

I am not sure whether this is actually public data. Following the endpoint URL, while **not** logged in, redirects me the OPH's starting page, not a SPARQL endpoint. The only thing that seems to be public is the inventory of processes, but as far as I see this does not provide any linked data.

config/migrations/add-decide-datasets/20260108152100-overview.ttl
```diff
@@ -0,0 +1,520 @@
...
+    dcat:servesDataset dataset:f7a4bbb3-1fef-4975-9049-ee933cfce30e
+.
+
+##############################################################################
+# Open Proces Huis
+
+dataset:74c12f17-f0eb-4bf2-8390-0d21655f00e3 a dcat:Dataset;
+    mu:uuid "74c12f17-f0eb-4bf2-8390-0d21655f00e3";
+    dct:title "Open Proces Huis procesdata"@nl;
+    dct:description "Proces-, diagram- en invatarisdata van Open Proces Huis"@nl;
```
mirdono: **Nitpick:** Small typo here, "inv**a**tarisdata" should be "inv**en**tarisdata".

Same for the `endpointDescription` in the corresponding `DataService` below.

config/migrations/add-decide-datasets/20260108152100-overview.ttl
```diff
@@ -0,0 +1,520 @@
...
+
+##############################################################################
+# Besluitdata
+
+service:0993dd37-87d7-4c20-894b-f7e3ffc87b7c a dcat:DataService ;
+    mu:uuid "0993dd37-87d7-4c20-894b-f7e3ffc87b7c";
+    dct:title "Internal SPARQL endpoint"@en, "Intern SPARQL endpoint"@nl;
+    dct:conformsTo <https://www.w3.org/TR/sparql11-protocol/> ;
+    dcat:endpointDescription "Internally available triple store for DECIDe"@en;
+    dcat:endpointURL "http://database:8890/sparql"^^xsd:anyURI ;
```
mirdono: **Question:** Is it fair to assume we will replace this later with a (public) sparql endpoint?

This `dcat:endpointURL`, as well as the same `dcat:acccessUrl`s below, are not web-resolvable IRIs. This IRI is, I think, only meaningful within a (local) development setting. Is this meant as a final value or will it be replaced with a (public) sparql endpoint when (or something else) later in the project?

config/migrations/add-decide-datasets/20260108152100-overview.ttl
```diff
@@ -0,0 +1,520 @@
...
+    dct:publisher agent:ABB;
+    dcat:contactPoint agent:7cb46783-0ca2-41a2-b28c-248b99e752e4;
+    dcat:theme <http://vocab.belgif.be/auth/datatheme/GOVE>;
+    dct:accessRights <http://publications.europa.eu/resource/authority/access-right/RESTRICTED>;
+    prov:alternateOf dataset:cf4d74f5-f39b-4193-a79c-6b46c8e3c9a6;
+    dcat:distribution distribution:7ccb4d96-a5fe-466b-8d54-74dcd318db31
+.
+distribution:7ccb4d96-a5fe-466b-8d54-74dcd318db31 a dcat:Distribution;
+    mu:uuid "7ccb4d96-a5fe-466b-8d54-74dcd318db31";
+    dcat:accessURL "http://database:8890/sparqlD"^^xsd:anyURI;
```
mirdono: **Nitpick:** A typo with a superfluous "D" at the end of the URL.

config/migrations/add-decide-datasets/20260108152100-overview.ttl
```diff
@@ -0,0 +1,520 @@
...
+    dcat:accessService service:0993dd37-87d7-4c20-894b-f7e3ffc87b7c;
+    rdfs:comment "Dit is rechtstraaks overgenomen van Lokaal Beslist harvester 1. De graph is enkel bereikbaar door ons op onze server (niet-publiek), maar de orginele Lokaal Beslist data is publiek. Data is opgeslagen in een graph (http://mu.semte.ch/graphs/oslo-decisions/landing) in triplestore op onze dev server"@nl
```
mirdono: **Nitpick:** Typo, I assume "rechtstraaks" should be "rechtstreeks".

config/migrations/add-decide-datasets/20260108152100-overview.ttl
```diff
@@ -0,0 +1,520 @@
...
+    dct:title "SPARQL endpoint DEV Decide"@en;
+    dct:description "SPARQL endpoint on the Dev server of Decide"@en;
+    dcat:accessURL "http://database:8890/sparql"^^xsd:anyURI; # only internally available
+    dcat:accessService service:0993dd37-87d7-4c20-894b-f7e3ffc87b7c;
+    dct:accessRights <http://publications.europa.eu/resource/authority/access-right/PUBLIC>;
+    dct:format <http://publications.europa.eu/resource/authority/file-type/SPARQLQRES>
+.
+
+
+distribution:743d147f-7079-4425-aa5b-0351dde216f4 a dcat:Distribution;
```
mirdono, line 475: **Question:** Is it intentional that this distribution lacks `accessRights`?

This distribution, and those below, do not specify `dcat:accessRights`, in contrast to previous distributions. Is this intentional?

config/migrations/add-decide-datasets/20260108152100-overview.ttl
```diff
@@ -0,0 +1,520 @@
...
+    dcat:theme <http://vocab.belgif.be/auth/datatheme/GOVE>;
+    dct:accessRights <http://publications.europa.eu/resource/authority/access-right/RESTRICTED>;
+    rdfs:seeAlso "https://app.gitbook.com/o/-MP9Yduzf5xu7wIebqPG/s/grRYjcouX0uZY2P6PYBl/ai-analysis-research/ai-approach-uc-0.0/named-entity-recognition"^^xsd:anyURI;
+    dcat:distribution distribution:db4e1b76-1a09-45fa-99be-32ddc5840a37
+.
+distribution:db4e1b76-1a09-45fa-99be-32ddc5840a37 a dcat:Distribution;
+    mu:uuid "db4e1b76-1a09-45fa-99be-32ddc5840a37";
+    dct:title "Freiburg PDFs on local decisions"@en;
+    dct:description "Freiburg PDFs on local decisions"@en;
+    rdfs:comment "Only internally available."@en;
```
mirdono: **Question** Any reason this comment is not expressed as a `RESTRICTED` value for `accessRights`?

config/migrations/add-decide-datasets/20260108152100-overview.ttl
```diff
@@ -0,0 +1,520 @@
...
+    dcat:theme <http://vocab.belgif.be/auth/datatheme/GOVE>;
+    dct:accessRights <http://publications.europa.eu/resource/authority/access-right/RESTRICTED>;
+    dcat:distribution distribution:7661627c-3cae-4991-bbdc-1f0a3f633781
+
+.
+distribution:7661627c-3cae-4991-bbdc-1f0a3f633781 a dcat:Distribution;
+    mu:uuid "7661627c-3cae-4991-bbdc-1f0a3f633781";
+    dct:title "Bamberg PDFs on local decisions"@en;
+    dct:description "Bamberg PDFs on local decisions."@en;
+    dcat:accessURL "https://u486391-sub1.your-storagebox.de/home/datasets/bamberg/bamberg_pdfs"^^xsd:anyURI; # only internally available
```
mirdono: **Question:** Should the code comment become a `dcat:accessRights` triple with a `RESTRICTED` value?

config/migrations/add-decide-datasets/20260108152100-overview.ttl
```diff
@@ -0,0 +1,520 @@
...
+
+##############################################################################
+# Publishers / contact points
+
+agent:7365f73a-b57c-4999-aa8b-61464d9d3c6f a foaf:Agent;
+    vcard:hasEmail "zinfo@more-rubin.de";
+    vcard:hasName "more! software GmbH"
+.
+
+agent:658900ce-63cf-4d45-8fd5-a67ed8deab2c a foaf:Agent;
```
mirdono, line 32: **Question:** Will we extend these with actual contact information later on?

Merely a name seems a bit insufficient as information contact point. Do we currently not have more information?

config/migrations/add-decide-datasets/20260108152100-overview.ttl
niels-triply: You're right, thanks for spotting this.

config/migrations/add-decide-datasets/20260108152100-overview.ttl
niels-triply: Thanks! That was blindly copied.

config/migrations/add-decide-datasets/20260108152100-overview.ttl
niels-triply: Yes, that was the intention. However, upon further reflection, I think it's better to just leave it out entirely (since then SHACL validation can remind us of the data quality issue later). I have done so.

config/migrations/add-decide-datasets/20260108152100-overview.ttl
niels-triply: It is intentional, since DCAT mandates putting `dcat:accessRights` on `dcat:CatalogedResource` (ie. datasets and data services) but not on distributions. See [DCAT](https://www.w3.org/TR/vocab-dcat-3/#Property:resource_access_rights) and [DCAT-AP](https://semiceu.github.io/DCAT-AP/releases/3.0.0/#controlled-vocabularies-to-be-used). What was not intentional is the stray `dcat:accessRights` on a distribution. I have removed it. Thanks!

* * *

There is a discussion to be had here, since I do think that `dcat:accessRights` make more sense on distributions than on datasets (and indeed there are some cases for us where some distributions of a public dataset are nevertheless restricted). My interpretation is that the access rights refer not to the concrete login restrictions of a particular distribution, but rather to the judgment on who should be able to access the dataset in principle. I'll also ask the others.

config/migrations/add-decide-datasets/20260108152100-overview.ttl
niels-triply: Same reason as above. Perhaps the comment was superfluous.

config/migrations/add-decide-datasets/20260108152100-overview.ttl
niels-triply: Same as above again!

config/migrations/add-decide-datasets/20260108152100-overview.ttl
niels-triply: I agree that the information is lacking (and in fact DCAT-AP-VL mandates the inclusion of an e-mail address). When collecting the datasets, I did ask for emails but there were few.

I've added information on Freiburg, LB and OPH by looking them up. As for datasets by DECIDe itself, perhaps we can put a generic e-mail address? Will also ask.

mirdono:
Two more things popped into my head (sorry, such things seem to randomly happen sometimes :joy:):

- Do not forget to add a `.graph` file so that the data is inserted in the correct graph. Currently the PR lacks such a file. (Clearly I had not yet tried to actually run the migration :blush:)
- This might also be a good moment to remove the previously added example data, more specifically that added in [#12](https://github.com/lblod/app-decide/pull/12).

niels-triply:
Haha, thanks, I keep remembering and then promptly forgetting the graph file again.

As for the example data, is it fine to simply remove the migration file or is it always good practice to explicitly delete it through another migration?

In any case, thanks for the review :)

mirdono:
> Haha, thanks, I keep remembering and then promptly forgetting the graph file again.
> 

I tend to use [mu-cli](https://github.com/mu-semtech/mu-cli) to generate the necessary files for migrations. Once installed locally, you can do `mu migrations new ttl NAME` in an app with a `migrations` service to generate the appropriate files in `config/migration/`. (That command will execute a [script](https://github.com/mu-semtech/mu-migrations-service/blob/master/scripts/new/run.sh) that is part of the mu-migrations-service.)

> As for the example data, is it fine to simply remove the migration file or is it always good practice to explicitly delete it through another migration?

The migrations from #12 have already been executed on the DEV environment. So the inserted data will have to be explicitly deleted via a new migration.
In general I tend to avoid removing migrations once they have been added to the app's primary branch. Exceptions would be when a migration turns out to be faulty or buggy (e.g. I have had migrations that work perfectly on a DEV environment, but fail on QA or PROD), or when contain sensitive data that should not be in the repository.

niels-triply:
That makes sense.

- I have added a migration that removes mock data.
- I have added better publishers. The e-mail for our own published datasets might still change depending on what Daan & Stefanie decide tomorrow, but that's not a change that needs further review :P 
- I have added a graph.
- I decided against using `xsd:anyURI` because it tripped up the resources service.
- I realized that I hadn't added the overarching catalog yet, so I have done so now.
- I didn't get a consensus on what accessRights really should mean, so I suggest to follow DCAT to the letter and put accessRights only on datasets and data services.

mirdono (approved):
Looks good.

Due to this PR I learned that ttl allows sparql-style prefix declarations. My apologies for badgering you about that in previous PRs.

niels-triply:
The fact that parsers do accept SPARQL-style prefixes is the reason I keep forgetting (because it won't complain), but in general it could still be argued that it's good practice to use @-notation for now... See https://www.w3.org/TR/turtle/#prefixed-name

Anyway, thanks!

niels-triply:
I will merge this and change the e-mail address to something less generic when we have one.

PR #44, opened 2026-02-03
Codelist matching (backend)

niels-triply:
These are the backend changes necessary for the addition of codelist matching.

Related PRs:
- https://github.com/lblod/frontend-harvesting-self-service/pull/42 for the frontend


niels-triply:
Hey @mirdono, Brecht already looked at the [frontend](https://github.com/lblod/frontend-harvesting-self-service/pull/42) and if those are the extent of the remarks, I expect the PR to be resolved.

But could you also check whether these backend changes make sense? In particular [bb10649](https://github.com/lblod/app-decide/pull/44/commits/bb1064927b9d5fa137954f252505cb561e1e5c34), since that adds ODRL policies for `sh:NodeShape`s. This is necessary because the decisions-to-be-mapped are described as `sh:NodeShape`s (so that more sophisticated filters on decisions can be implemented later).

mirdono:
The changes concerning the (ODRL) authorisation policy look correct to me. I was unable to explicitly test it with the `fronted-harvesting` as I cannot select codelist mapping as operation for jobs. Not sure if that is expected at this point or an issue with my local stack.

Thanks for taking the time to already update the ODRL policy!

brechtvdv:
Who can solve these conflicts?

mirdono:
> Who can solve these conflicts?

Sorry, I caused those with #48. Will fix them.

mirdono:
> > Who can solve these conflicts?
> 
> Sorry, I caused those with #48. Will fix them.
> 

And done. Note I had to rebase onto `development` and force-push the resulting commits.

The only commit that changed content-wise is `ce34a27 Add authorization for nodeshapes.`:
- I regenerated the `config.nt` and `decideAuthorizationPolicy.lisp`. Especially `config.nt` tends to cause conflicts as the order of triples in the file can vary strongly between different generations.
- This commit previously also included adding the `:sh` prefix in `decideAuthorizationPolicy.lisp`, since that prefix was already defined this addition is not necessary. (Did not notice that when reviewing earlier. I assume the file was previously edited manually instead of regenerated using the `odrl-parser` service.)

PR #50, opened 2026-02-10
feat: configuration for ACM/IDM authentication

mirdono:
This PR configures the DECIDe app for integration with the ACM/IDM authentication flow.

### Dispatcher
Update service's last `sessions` route to forward to the `acmidm-login` service instead of the `login` service. The latter should only be used for the plain username/password login used for the dashboard, which is captured by rules earlier in the configuration.

### Sparql-parser
Define additional access rights necessary for the frontend to properly handle authentication and retrieving user information.

The `http://mu.semte.ch/graphs/organizations/UUID` graphs contain account information for users associated with that particular organisation. The frontend queries for the account belonging to a session when it tries to load the data for the current session. A user only gets read access to the organisation's graph its account is associated with.

Similarly, the `http://mu.semte.ch/graphs/bestuurseenheden-bestuursorganen` graph contains data on administrative units. The frontend will also query for this when loading a current session. As this should only contain public data I opted to grant public read access to this graph.

### Add sink service
Add a `sink` service, this service was already configure as `LOG_SINK_URL` for the `acmidm-login` service. But because the service was never actually added to the stack this caused the `acmidm-login` service to crash when it tried to contact it. The `sink` service allow `app-http-logger` to log more complete information.

## Related PRs
- [Frontend PR](https://github.com/lblod/frontend-decide/pull/13)

## Related tickets
- LBRON-838

### DIFF
--- a/config/authorization/decide.lisp
+++ b/config/authorization/decide.lisp
@@ -1,6 +1,7 @@
 (in-package :acl)
 
 (define-prefixes
+  :adms "http://www.w3.org/ns/adms#"
   :besluit "http://data.vlaanderen.be/ns/besluit#"
   :cms "http://mu.semte.ch/vocabulary/cms/"
   :cogs "http://vocab.deri.ie/cogs#"
@@ -129,6 +130,11 @@
   ("tasks:CronSchedule" -> _ )
   ("schema:repeatFrequency" -> _ ))
 
+(define-graph organization ("http://mu.semte.ch/graphs/organizations/")
+  ("foaf:Person" -> _)
+  ("foaf:OnlineAccount" -> _)
+  ("adms:Identifier" -> _))
+
 (supply-allowed-group "public")
 
 (grant (read)
@@ -156,3 +162,15 @@
       SELECT DISTINCT ?account WHERE {
       <SESSION_ID> session:account ?account.
       }")
+
+(supply-allowed-group "organization-member"
+  :parameters ("session_group")
+  :query "PREFIX ext: <http://mu.semte.ch/vocabularies/ext/>
+          PREFIX mu: <http://mu.semte.ch/vocabularies/core/>
+          SELECT ?session_group ?session_role WHERE {
+            <SESSION_ID> ext:sessionGroup/mu:uuid ?session_group.
+          }")
+
+(grant (read)
+       :to-graph organization
+       :for-allowed-group "organization-member")
--- a/config/authorization/decide.lisp
+++ b/config/authorization/decide.lisp
@@ -174,3 +174,14 @@
 (grant (read)
        :to-graph organization
        :for-allowed-group "organization-member")
+
+;; TODO: Is this supposed to be the final graph?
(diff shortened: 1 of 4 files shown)


Rahien (approved):

config/authorization/decide.lisp
```diff
@@ -137,3 +143,26 @@
...
+
+(grant (read)
+       :to-graph organization
+       :for-allowed-group "organization-member")
+
+;; TODO: Is this supposed to be the final graph?
+;; NOTE (10/02/2026): Graphs <http://mu.semte.ch/graphs/organizations> does NOT contain the
+;; `besluit:Bestuurseenheid' type (only `org:Organization'), used that one causes resource service
+;; to bug out
+(define-graph organizations ("http://mu.semte.ch/graphs/bestuurseenheden-bestuursorganen")
```
mirdono, line 182: **Question:** Is this be the final graph in which this data is located?

I am not certain whether this is the final graph in which this data will remain or whether this is a temporary graph for one of the pipelines. Note, the `acmidm-login` service will use the same graph as its `MU_APPLICATION_GRAPH`.

MartijnBogaert: Instead of adding a new pipeline, we decided to add bestuurseenheden and bestuursorganen through a migration (the use of pipelines is already demonstrated abundantly, so this was just easier and quicker). The migration puts its data in that `http://mu.semte.ch/graphs/bestuurseenheden-bestuursorganen` graph indeed (cfr. https://github.com/lblod/app-decide/blob/development/config/migrations/bestuurseenheden-and-organizations-flanders/20260127153400-add-bestuurseenheden-bestuursorganen.graph).
There are also two other migrations that transform the besuurseenheden and bestuursorganen to org organizations. That data ends up in `http://mu.semte.ch/graphs/organizations`. (cfr. https://github.com/lblod/app-decide/blob/development/config/migrations/bestuurseenheden-and-organizations-flanders/20260128143300-transform-bestuurseenheden-to-organizations.sparql and https://github.com/lblod/app-decide/blob/development/config/migrations/bestuurseenheden-and-organizations-flanders/20260128144300-transform-bestuursorganen-to-organizations.sparql).
If these target graphs are inconvenient, I can update them, but that will entail some data moving.

mirdono: Thanks, that clarifies it for me. Given the context of this app a one-time import indeed makes sense compared to setting up a consumer or something.

I am fine with using the `http://mu.semte.ch/graphs/bestuurseenheden-bestuursorganen`, we just need to keep in mind we are actually using this data and do not delete it at some later point. 

(An alternative would be to further modify our frontend and resources config to use `org:Organization` instead of `besluit:Bestuurseenheid`. But that is probably going to be messy to work alongside mock-login which, I think, relies on the latter type. And in any case the `acmidm-login-service` will still expect `besluit:Bestuurseenheid` resources. Imo modifying that service will takes us too far.)

Rahien (approved):

PR #63, opened 2026-02-27
split compose file into logical subunits

Rahien:
uses `include` to split the main compose file into different files according to logical 'coherence' to provide a better overview of what is going on


compose/gent.yml
```diff
@@ -0,0 +1,51 @@
+x-logging: &default-logging
```
mirdono: **Nitpick:** Rename file to something more generic.

I would rename this file something more generic like `oslo.yml` or `lblod.yml`. I **think** the only Ghent-specific thing here is the filtering of their decisions. Such naming would make it similar to `oparl.yml` where the fact that it is specific to Freiburg is a configuration thingie (in theory at least I do not know how generic the underlying data model is.)

compose/base-compose.yml
```diff
@@ -0,0 +1,105 @@
...
+    logging: *default-logging
+  sink:
+    image: nginx:1.15.2
+    volumes:
+      - ../config/sink/sink.conf:/etc/nginx/conf.d/default.conf
+    labels:
+      - 'logging=true'
+    restart: always
+    logging: *default-logging
+  frontend:
```
mirdono, line 92: **Comment:** Not sure whether this is part of the "base".

At least once the frontend is split its corresponding service should probably move to more specific files. For example, the frontend with the VC part should move to `data-space.yml`.

compose/login.yml
```diff
@@ -0,0 +1,36 @@
+x-logging: &default-logging
```
mirdono: **Nitpick:** Merge this into other files.

I am not sure whether it is worthwhile to put these two services separate. Imo the `acmidm-login` is tied to the `frontend` and should probably be put close to it. That would also make it easier to see the overlap in configuration of these services (although one would already see that in the override.) Similarly, I **think** that the `login` service is only relevant for the harvesting dashboard, so it would make sense to put those together.

mirdono (commented):
I like the idea and am definitely in favour. Left some minor non-blocking remarks for your consideration.

For future reference the [include documentation](https://docs.docker.com/reference/compose-file/include/), I was not yet familiar with this syntax.

compose/base-compose.yml
Rahien: Good point, but would it be fine to leave it in base until it is split? That way this one can proceed without conflicting with everything everyone else is doing in the app repo :joy:

MartijnBogaert (approved):
I'm a big fan of this way of working!

compose/base-compose.yml
mirdono: Yes, atm it makes sense to put it there since the current frontend is a mix of functionality. More something we need to keep in mind when splitting it up.

PR #64, opened 2026-03-10
Update organization config

brechtvdv:
Before this PR, organizations (following ORG-EP standard) resided in eli-misc.lisp, because it was not used.
But now the frontend dashboard requires looking up governing bodies, in the case of the PDF pipeline:
https://github.com/lblod/frontend-harvesting-self-service/pull/46

This PR does following things:
- Renames eli-misc.lisp to org.lisp, because organizations will now be used in frontend dashboard 
- Update org:classification datatype to :url
- Moves has-sub-organizations and sub-organization-of from has-one to has-many


mirdono:
Could you add some explanation to the PR, this should address questions like (a general list, so not all points will be (equally) applicable here):
- What does the PR actually aim to do/change/improve/...?
- Why is this change necessary?
- Why is the change implemented this way? Did you consider (and discard) alternative approaches? If so, which ones and why?
- How can a reviewer test the changes?
- Are there things to pay special attention to?

Also try to mention related tickets for easier traceability.

The lack of this kind of information unnecessarily complicates the review process. In the PR's current shape, a reviewer would have to try to reverse engineer such information from the actual changes and/or ask the PR's creator directly. Both of these require additional time and effort, and likely still lead to incomplete information. For example, alternative approaches that were considered and discarded are typically not documented in the code. Furthermore, the creator may not be available to answer questions.

Imo a PR should aim to make the review process as easy as possible. In other words, it should allow reviewers to focus on the essential complexity of assessing the actual change in all its facets. Not require reviewers to figure out, for instance, what goal the change is actually supposed to achieve. (Imo this typically requires more than providing sufficient description, but also includes having a clear commit history. But that is a whole other discussion.)

MartijnBogaert (approved):

PR #65, opened 2026-03-12
allow searching organizations and expressions based on owning-body

Rahien:
## Description
adds owning-body to search config, also link oslo bestuurseenheden and orgs(T) using org:hasSubOrganization

## How to Test

- `drc kill search elasticsearch migrations`
- `drc up -d migrations`
- `rm -rf ./data/elasticsearch && mkdir ./data/elasticsearch` to remove the existing search indexes so they are rebuilt
- `drc up -d migrations` wait for them to complete
- `drc up -d search elasticsearch` wait for search to be done creating indexes, this will take a while, fetch a snack

Now you can run this search query:
`POST http://localhost/search/organizations/large-search`
body:
```json
{
  "filter": {
    "owning-body": "http://data.lblod.info/id/bestuurseenheden/353234a365664e581db5c2f7cc07add2534b47b8e1ab87c821fc6e6365e6bef5",
    ":fuzzy:": "Gemeenteraad"
    
  }
}
```


config/migrations/20251104134600-add-eli-to-public.sparql
```diff
@@ -1,13 +0,0 @@
-INSERT {
```
mirdono, line 1: **Question:** Should any previously inserted data be removed as well?

Assuming this migration was previously executed, should we also remove the data it inserted? Or is it okay that this implicitly happens when we reset `dev` with the "clean" data from `test`?

config/search/config.json
```diff
@@ -1,12 +1,6 @@
 {
   "persist_indexes": true,
-  "eager_indexing_groups": [
-    [{ "variables": [], "name": "public" }],
-    [
-      { "variables": [], "name": "logged-in" },
```
mirdono, line 6: **Question:** Is this change related to the added index?

I am not clear on why this is necessary here. Is this merely a config cleanup you noticed while working on this or something?

mirdono (approved):
Left some non-blocking questions. Tested it locally and the example query returns the municipality council for Ghent, which I assume is the expected result as the Ghent administrative unit is given as input.

Thanks for also fixing the duplicate graph names in the authorisation policy. (Although I would have done in a separate PR as I think it is unrelated to the changes proposed here.)

config/migrations/20251104134600-add-eli-to-public.sparql
Rahien: indeed this is for clean resets of the data space and the only place where this has had any impact is on local machines (who need a reset anyway) and dev

config/search/config.json
Rahien: this is indeed cleanup for something we don't need that i noticed while working on this one. We don't have a logged in user that searches, only public searches. So pre-filling all indexes twice is a lot of effort for nothing

PR #66, opened 2026-03-13
Split frontends.

niels-triply:
A split has been made between the different front-ends:

- https://github.com/lblod/frontend-decide-dcat for the DCAT catalogue
- https://github.com/lblod/frontend-decide-yasgui for the SPARQL interface and necessary VC authorization
- https://github.com/lblod/frontend-decide-human-validator for the AI annotations

# Note on frontend PRs

This is the PR for the backend, and the changes are minimal. 

At the front-end, while figuring out how to pull versioned images for Woodpecker, I had already caught up the new repositories to `master`. The repositories are simply forks of https://github.com/lblod/frontend-decide (in the sense that I initially created them by simply adding extra remotes to that original repo). However, GitHub won't recognize them as forks that way.

I realize that makes the front-ends a bit harder to review. Apologies for that. All changes are relative to https://github.com/lblod/frontend-decide/commit/80806d0baeda3c620e479ff5fea2256e19e86e31. Would it help the review process to create 'mock' frontend PRs against `80806d0...`?

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,15 +105,6 @@ To show how decisions are linked with SDGs, a Policy impact report tool is being
 
 ## Pipeline dashboard
 
-### Accessing the dashboard from your local machine
-
-Since we use dispatcher v2, which dispatches on hostname, we'll have to update `/etc/hosts`.
-Add an entry similar to the following. Ensure the first part of the domain starts with `dashboard`.:
-
-```
-127.0.0.1 dashboard.localhost
-```
-
 ### Jobs
 
 There are two types of jobs: harvesting and scheduled. The harvesting job is a one-time run of a job, while the scheduled job is triggered periodically following a cron pattern.
@@ -121,3 +112,16 @@ There are two types of jobs: harvesting and scheduled. The harvesting job is a o
 By pressing "Create new job", a job type ("operation") can be selected to create a new job.
 
 ![alt text](doc/dashboard-create-new-2.png)
+
+## Frontends
+
+### Accessing the frontends from your local machine
+
+We use dispatcher v2, which dispatches different frontends based on hostname. If this does not work out of the box, you may have to add an entry similar to the following to your `/etc/hosts`:
+
+```
+127.0.0.1 dashboard.localhost
+127.0.0.1 dcat.localhost
+127.0.0.1 human-validator.localhost
+127.0.0.1 yasgui.localhost
+```
--- a/compose/base-compose.yml
+++ b/compose/base-compose.yml
@@ -89,8 +89,8 @@ services:
       - 'logging=true'
     restart: always
     logging: *default-logging
-  frontend:
-    image: lblod/frontend-decide:0.0.3
+  frontend-yasgui:
+    image: lblod/frontend-decide-yasgui:0.0.4
     environment:
       # The following are environment-specific and should be set in the
       # environment's appropriate override file.
@@ -99,6 +99,18 @@ services:
       EMBER_OAUTH_API_BASE_URL: '<OAUTH_API_BASE_URL>'
       EMBER_OAUTH_API_LOGOUT_URL: '<OAUTH_API_LOGOUT_URL'
       EMBER_OAUTH_API_SCOPE: '<OAUTH_API_SCOPE>'
+    labels:
+      - 'logging=true'
+    restart: always
+    logging: *default-logging
+  frontend-human-validator:
+    image: lblod/frontend-decide-human-validator:0.0.4
+    labels:
+      - 'logging=true'
+    restart: always
+    logging: *default-logging
+  frontend-dcat:
+    image: lblod/frontend-decide-dcat:0.0.4
     labels:
       - 'logging=true'
     restart: always
--- a/config/dispatcher/dispatcher.ex
+++ b/config/dispatcher/dispatcher.ex
@@ -65,6 +65,10 @@ defmodule Dispatcher do
     Proxy.forward conn, path, "http://question-answering/uc2/"
   end
 
+  match "/api/sparql", %{ accept: [:any], layer: :api_services, reverse_host: ["yasgui" | _rest]} do
+    Proxy.forward conn, [], "http://database:8890/sparql"
+  end
+
   match "/api/sparql", %{ accept: [:any], layer: :api_services } do
     Proxy.forward conn, [], "http://database:8890/sparql"
   end
@@ -313,6 +317,14 @@ defmodule Dispatcher do
     Proxy.forward(conn, path, "http://mocklogin/sessions/")
   end
 
+  match "/sessions/*path", %{reverse_host: ["yasgui" | _rest]} do
+    Proxy.forward(conn, path, "http://login/sessions/")
+  end
+
+  match "/mock/sessions/*path", %{reverse_host: ["yasgui" | _rest]} do
+    Proxy.forward(conn, path, "http://mocklogin/sessions/")
+  end
+
   match "/sessions/*path", %{layer: :api_services, accept: %{any: true}} do
     Proxy.forward(conn, path, "http://acmidm-login/sessions/")
   end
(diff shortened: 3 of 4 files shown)


config/dispatcher/dispatcher.ex
```diff
@@ -339,31 +347,80 @@ defmodule Dispatcher do
   end
 
   match "/index.html", %{layer: :static} do
-    forward(conn, [], "http://frontend/index.html")
+    forward(conn, [], "http://frontend-dcat/index.html")
```
mirdono: **Issue:** This rule shadows similar subsequent ~index.html~ rules in the configuration.

The rules in the configuration are tried in order of top to bottom. This one will match any requests for the `/index.html` route, irrelevant of the `reverse_host` set in the rules below. It should be removed so that the added rules are actually used.

config/dispatcher/dispatcher.ex
```diff
@@ -339,31 +347,80 @@ defmodule Dispatcher do
...
   end
 
   get "/assets/*path", %{layer: :static} do
-    forward(conn, path, "http://frontend/assets/")
+    forward(conn, path, "http://frontend-dcat/assets/")
```
mirdono: **Issue:** Same as comment on `index.html` rule.

This is also the rule that causes the the errors concerning the asset hashes for the human-validator and yasgui frontends. Due to this rule it is actually trying to retrieve the dcat frontend assets in them.

config/dispatcher/dispatcher.ex
```diff
@@ -339,31 +347,80 @@ defmodule Dispatcher do
...
   end
 
   get "/@appuniversum/*path", %{layer: :static} do
-    forward(conn, path, "http://frontend/@appuniversum/")
+    forward(conn, path, "http://frontend-dcat/@appuniversum/")
```
mirdono: **Issue:** Same as comment on `index.html` rule.

mirdono (changes requested):
There are superfluous rules in the dispatcher configuration, causing some of the newly added rules to be shadowed. See inline comments.

**Comment:** Also update the information in the frontend's `package[-lock].json` files.

Some fields in these files still contain the values of the, now, old frontend: `name`, `description`, `version`, `repository`.

mirdono:
Note, I'll leave comments concerning the frontends here since there is no convenient other place[^1]

**Issue:** The yasgui frontend does not work anymore with mock-login.

I logged in as Jane Doe via mock-login. When I try to execute a query (actually just executing the filled in example query) I get an error:

``` html
Method Not Allowed (#405)

<html>
<head><title>405 Not Allowed</title></head>
<body>
<center><h1>405 Not Allowed</h1></center>
<hr><center>nginx/1.19.6</center>
</body>
</html>
```

Console contains the following error (next to HTTPS-Only and cors error that the original frontend already throws):

``` shell
Error: Method Not Allowed
    b http://yasgui.localhost/assets/chunk.420.70076fcad097a86e32f7.js:3354
    emit http://yasgui.localhost/assets/chunk.420.70076fcad097a86e32f7.js:7092
    onreadystatechange http://yasgui.localhost/assets/chunk.420.70076fcad097a86e32f7.js:3370
chunk.420.70076fcad097a86e32f7.js:4103:43
```

Doing the same thing using the old `frontend-decide` (running locally with `eds --proxy http://host/`) does work as expected.


[^1]: Personal opinion: When creating a new project/repository I tend to populate the primary branch (main, master, development) with some placeholder content (README, some skeleton structure, ...) and do the initial implementation in a different branch. That way you can make a PR against the almost empty primary branch which contains all added code. In this case that would have made it easier to see what remains, instead of trying to piece that together manually. (Possibly my diff skill are insufficient here.) That said, I do not think it is worth it to spend time on doing this here since the work of splitting etc. has already been done.

config/dispatcher/dispatcher.ex
mirdono: Some nuance with respect to my top to bottom remark here (insight resulted from our discussion on chat.) Since we use layers, rules are tried top to bottom **per layer**.

niels-triply:
Thanks a lot for the thoughtful review @mirdono.

- Updated the `package{,-lock}.json` for each frontend.
- Added instructions to update `/etc/hosts` in the `README.md`.
- Reordered the dispatcher rules to fix the hash issues.
- Renamed `reverse_host` `validate`  to `human-validator` for consistency.
- **Question:** I started the version from `0.0.4` because the three frontends split off from the original repo at `0.0.3`. On second thought, just starting from `0.0.1` makes more sense. Is that worth it, and if so, would it confuse Woodpecker?

### SPARQL issue

Consider that we used to have:

```elixir
  define_layers([:static, :sparql, :api_services, :frontend_fallback, :resources, :not_found])
  # ...
  match "/api/sparql", %{ reverse_host: ["yasgui" | _rest], accept: [:any], layer: :sparql } do
    Proxy.forward conn, [], "http://database:8890/sparql"
  end
  # ...
  match "/*_path", %{reverse_host: ["yasgui" | _rest], accept: %{html: true} } do
    forward(conn, [], "http://frontend-yasgui/index.html")
  end
```

It seems that the `/*_path` rule takes precedence over `/api/sparql`... *unless* `/api/sparql` is either in the `:static` layer or has no layer at all. To address this, in a73edaafb859bb5e5785db2bd73eead348825deb I add a `:frontend` layer for the `/*_path` rules that explicitly comes after `:sparql` but before `:api_services`. This solves the issue.

(That said, I'm still not *entirely* sure of the behaviour of the dispatcher. It seems that there is some interaction between `reverse_host` and the execution order, such that a rule without layers can come after some `:static` rules but before some `:api_services` rules. @Rahien  is this okay?)


### Branches

> When creating a new project/repository I tend to populate the primary branch [...] and do the initial implementation in a different branch. 

I completely agree and this actually was my approach... but then I forgot about it to test Woodpecker builds. Sorry! Note however that you can do ad-hoc diffs within GitHub: <https://github.com/lblod/frontend-decide-human-validator/compare/80806d0baeda3c620e479ff5fea2256e19e86e31...master>

README.md
```diff
@@ -105,19 +102,24 @@ To show how decisions are linked with SDGs, a Policy impact report tool is being
...
 
 There are two types of jobs: harvesting and scheduled. The harvesting job is a one-time run of a job, while the scheduled job is triggered periodically following a cron pattern.
 
 By pressing "Create new job", a job type ("operation") can be selected to create a new job.
 
 ![alt text](doc/dashboard-create-new-2.png)
+
+## Frontends
+
+### Accessing the frontends from your local machine
```
mirdono, line 118: **Comment:** Whether this is necessary depends on your host machine.

I, for instance, did not have to edit my `/etc/hosts` file to be able to access the frontends, it just worked out of the box (after resolving the hash issue.) I do not know what determines whether this is necessary. Maybe add a sentence that this step may not be necessary on ones system.

compose/base-compose.yml
```diff
@@ -16,6 +16,9 @@ services:
     logging: *default-logging
   dispatcher:
     image: semtech/mu-dispatcher:2.1.0-beta.2
+    environment:
+      LOG_LAYER_START_PROCESSING: true
+      LOG_LAYER_MATCHING: true
```
mirdono: **Nitpick:** I would keep this out of the versioned configuration.

These are useful for debugging, as we noticed while digging into the layer issue. But I do not think we need to enable on for the deployed instances of the app.

mirdono (commented):
Left some non-blocking comments on the latest changes. As far as I tested everything works as expected and the individual frontends seem in order I did not check every line, but at first sight the unnecessary parts seem to be removed.

Note, once this PR is merged we should probably archived the now deprecated `frontend-decide` repository.

README.md
niels-triply: Good point. I'm finding conflicting information on whether `*.localhost` is always accessible. I've changed the message slightly.

compose/base-compose.yml
niels-triply: Sorry, that was by-catch. Fixed!

mirdono:
> - **Question:** I started the version from `0.0.4` because the three frontends split off from the original repo at `0.0.3`. On second thought, just starting from `0.0.1` makes more sense. Is that worth it, and if so, would it confuse Woodpecker?

Imo there is an argument to be made for either option. I would keep it as is with `0.0.4` since that is already published on our dockerhub (e.g. [dcat-frontend](https://hub.docker.com/r/lblod/frontend-decide-dcat/tags)). Changing it to `0.0.1` is just going to be confusing unless we also remove the `0.0.4` (which is extra work that probably requires involving someone with more access rights.)

> ### SPARQL issue
> 
> Consider that we used to have:
> 
> ```elixir
>   define_layers([:static, :sparql, :api_services, :frontend_fallback, :resources, :not_found])
>   # ...
>   match "/api/sparql", %{ reverse_host: ["yasgui" | _rest], accept: [:any], layer: :sparql } do
>     Proxy.forward conn, [], "http://database:8890/sparql"
>   end
>   # ...
>   match "/*_path", %{reverse_host: ["yasgui" | _rest], accept: %{html: true} } do
>     forward(conn, [], "http://frontend-yasgui/index.html")
>   end
> ```
> 
> It seems that the "/*_path" rule takes precedence over `/api/sparql`... *unless* `/api/sparql` is either in the `:static` layer or has no layer at all. To address this, in https://github.com/lblod/app-decide/commit/7808df16e5ff6142564e9a7deb5c7a0f495d682e I add a `:frontend` layer for the `/*_path` rules that explicitly comes after `:sparql` but before `:api_services`. This solves the issue.
> 
> (That said, I'm still not *entirely* sure of the behaviour of the dispatcher. It seems that there is some interaction between `reverse_host` and the execution order, such that a rule without layers can come after some `:static` rules but before some `:api_services` rules. @karel.kremer is this okay?)

@Rahien is going to work better to ping him :-)
> 
> ### Branches
> 
> > When creating a new project/repository I tend to populate the primary branch [...] and do the initial implementation in a different branch. 
> 
> I completely agree and this actually was my approach... but then I forgot about it to test Woodpecker builds. Sorry! Note however that you can do ad-hoc diffs within GitHub: <https://github.com/lblod/frontend-decide-human-validator/compare/80806d0baeda3c620e479ff5fea2256e19e86e31...master>
> 

Yeah, I did something similar locally. But that only shows you what was removed and modified, which are indeed the changes that happened in your commits. But in this particular case I would be more interested in seeing what remains after those changes. The diff from a PR against an empty branch would provide me exactly that. But don't worry about it, was able to review it nonetheless :-)

mirdono (approved):
As far as I am concerned this is okay now. Unless all-knowing Rahien has objections wrt the layer thingie (or something else), we are good to go.

PR #71, opened 2026-03-20
Process frontend only after API services.

niels-triply:
Without this change, the `/*_path` rule on host `yasgui.*` matches before the VC rules, which means that verifiable credentials no longer work.


Rahien:
fixed in later changes

PR #73, opened 2026-03-25
keep ds for the dataspace (dcat) part of the app

Rahien:
we already communicated ds as a subdomain for dcat, let's keep it


mirdono:
**Comment:** Also update `Frontends` entry in README.

The [README](https://github.com/lblod/app-decide/blob/karel/keep-ds/README.md?plain=1#L124) also mentions the `dcat` subdomain, best to update this to avoid future confusion.

mirdono (approved):
Looks good

PR #72, opened 2026-03-25
add annotaiton-review service so we can list decisions

Rahien:
see https://github.com/lblod/frontend-decide-human-validator/pull/1 for how to test

PR #41, opened 2026-01-26
feat: use ODRL policy for authorisation configuration

mirdono:
Use an ODRL authorisation policy as configuration for `sparql-parser`. With support for ODRL configuration being added to `sparql-parser` we no longer need the `odrl-parser-service` to translate an ODRL policy to `sparql-parser`'s lisp ACL.

- Update the app's `database` service configuration to use ODRL as implemented in [sparql-parser](https://github.com/mu-semtech/sparql-parser/pull/12) using builds from a feature-branch.
- Remove the `odrl-parser` service and its configuration.
- Remove the project script to convert ttl files to n-triples files.

### Running the app with ODRL configuration
1. Checkout the branch for this PR.
2. Pull the feature build for the `database` service: `docker compose pull database`
3. (Re)Up `database` service using `docker compose up -d database`, or start the app's stack as usual if not yet running.
4. The `database` service logs should now contain the following lines indicating it read its configuration from the ttl file.

``` shell
database-1  |  >> Loading configuration via ODRL
database-1  |  >> INFO: Reading ODRL policy from ../config/config.ttl
```

5. The app should work the same as before.

### Understanding and modifying the ODRL policy
As part of this feature `sparql-parser`'s README has been extended with a [how-to guide for ODRL policies](https://github.com/mu-semtech/sparql-parser/tree/feature/odrl-configuration?tab=readme-ov-file#defining-an-authorization-policy-in-odrl). Feedback can be provided as part of [PR #12](https://github.com/mu-semtech/sparql-parser/pull/12) for sparql-parser.

### Check policy equivalence
To double-check that the ODRL policy and the original lisp configuration are equivalent. The policy update [commit](https://github.com/lblod/app-decide/commit/9ad36eb1c40e97573c4b59eabee2486d033013d0) contains both the original lisp config `config/authorization/decide.lisp` and the one generated from the ttl file `config/odrl-parser/decideAuthorizationPolicy.lisp`. You can check out this commit and compare the files, but take into consideration things:
- The order of the elements in these two files differ, they cannot be simply compared line by line.
- The generated `decideAuthorizationPolicy` config defines more (unnecessary) prefixes as the `odrl-parser-service` simply adds all prefixes it knows to the output it generates.

[...]

PR #74, opened 2026-03-26
allow voting on annotations

Rahien:
mu-auth config to allow registering reviewannotations (which have both oa:Annotation and ext:ReviewAnnotation as a type)

### DIFF
--- a/config/authorization/decide.lisp
+++ b/config/authorization/decide.lisp
@@ -147,6 +147,11 @@
 (define-graph human-validation ("http://mu.semte.ch/graphs/public/human-validation/")
   ("ext:ReviewAnnotation" -> _))
 
+(define-graph ai ("http://mu.semte.ch/graphs/ai")
+  ("oa:Annotation" -> _)
+  ("oa:SpecificResource" -> _)
+  ("oa:TextPositionSelector" -> _))
+
 (supply-allowed-group "public")
 
 (grant (read)
@@ -167,6 +172,10 @@
 
 (grant (read)
        :to-graph harvested-pdf
+       :for-allowed-group "public")
+
+(grant (read)
+       :to-graph ai
        :for-allowed-group "public")
 
 (grant (read write)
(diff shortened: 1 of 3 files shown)


mirdono:
**Issue:** This branch lacks the `compose/validation.yml` file.

It looks like the missing file was added directly to `development`, and not to the branch of #72 from which this PR's branch was started from. I will continue reviewing with a rebased branch locally. But please take care to ensure PR branches are up to date with `development`, at least as far as possible when creating a PR/asking review on a PR. It is a bit annoying to have to deal with such things before you can actually get to the reviewing part.

config/authorization/decide.lisp
```diff
@@ -144,6 +144,9 @@
   ("foaf:OnlineAccount" -> _)
   ("adms:Identifier" -> _))
 
+(define-graph human-validation ("http://mu.semte.ch/graphs/public/human-validation/")
```
mirdono: **Remark:** This change should propagate to #41.

Depending on which PR gets merged first we should:
- update #41 with these policy changes; or
- update the ODRL policy instead of lisp-style in this PR.

config/authorization/decide.lisp
```diff
@@ -170,6 +173,41 @@
...
+;     :for "public"
+;   ))       
+
+; (with-scope "http://services.semantic.works/annotation-review-service"
+;   (grant (read)
+;     :to harvested-pdf
+;     :for "public"
+;   ))       
+;; instead we need to give public write access to user reviews
+(grant (read write)
```
mirdono, line 216: **Remark:** This change should propagate to #41.

Same as remark for `define-graph`.

config/authorization/decide.lisp
mirdono: **Comment:** The trailing "/" is not needed.

I think this is meant as the full graph URI, and what we will not append other segments to it.

mirdono (approved):

PR #79, opened 2026-04-07
add impact migration and some resource tweaks

Rahien:
see https://github.com/lblod/frontend-decide-human-validator/pull/11 for description and how to test

### DIFF
--- a/config/migrations/20260407111121-create-impact.sparql
+++ b/config/migrations/20260407111121-create-impact.sparql
@@ -1,0 +1,16 @@
+PREFIX skos: <http://www.w3.org/2004/02/skos/core#>
+PREFIX mu: <http://mu.semte.ch/vocabularies/core/>
+
+INSERT DATA {
+  GRAPH <http://mu.semte.ch/graphs/public> {
+    <http://mu.semte.ch/vocabularies/ext/impact/positive> a skos:Concept ;
+    skos:inScheme <http://mu.semte.ch/vocabularies/ext/impact> ;
+    mu:uuid "0e432b1b-87ad-4c32-9fc5-b151dd49f60d" ;
+    skos:prefLabel "Positive" .
+
+    <http://mu.semte.ch/vocabularies/ext/impact/negative> a skos:Concept ;
+    skos:inScheme <http://mu.semte.ch/vocabularies/ext/impact> ;
+    mu:uuid "1e122909-a685-4c0f-8b61-d639173b0a58" ;
+    skos:prefLabel "Negative" .
+  }
+}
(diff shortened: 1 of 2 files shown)


niels-triply (approved):
Looks good to me. I verified that the changes work and are reflected in the data; I cannot immediately check the front-end changes without also loading the annotations, but that is part of the front-end review anyway

config/migrations/20260407111121-create-impact.sparql
```diff
@@ -0,0 +1,16 @@
+PREFIX skos: <http://www.w3.org/2004/02/skos/core#>
+PREFIX mu: <http://mu.semte.ch/vocabularies/core/>
+
+INSERT DATA {
+  GRAPH <http://mu.semte.ch/graphs/public> {
+    <http://mu.semte.ch/vocabularies/ext/impact/positive> a skos:Concept ;
+    skos:inScheme <http://mu.semte.ch/vocabularies/ext/impact> ;
```
mirdono, line 11: **Question:** Should we define this scheme more explicitly as a `skos:ConceptScheme`.

Currently, no further information if specified for this concept scheme. Should we not define it at least as a `skos:ConceptScheme` and add a description and/or label so the purpose of this scheme is clearer?

config/migrations/20260407111121-create-impact.sparql
```diff
@@ -0,0 +1,16 @@
...
+    mu:uuid "0e432b1b-87ad-4c32-9fc5-b151dd49f60d" ;
+    skos:prefLabel "Positive" .
+
+    <http://mu.semte.ch/vocabularies/ext/impact/negative> a skos:Concept ;
+    skos:inScheme <http://mu.semte.ch/vocabularies/ext/impact> ;
+    mu:uuid "1e122909-a685-4c0f-8b61-d639173b0a58" ;
+    skos:prefLabel "Negative" .
+  }
```
mirdono, line 19: **Question:** Do we also need a concept to specify "no impact"?

When a decision has no impact (for an SDG) this is currently implied in having no linked impact concept. Bur should we not have this more explicitly with something like a "no impact", or "neutral", concept? That would allow us to differentiate between decisions that we know have no impact and those for which we do not know whether they have impact. (Assuming the AI services creating this data can differentiate such things.)
(This feels, at least, related to discussions in the past. But I cannot recall whether we decided something on this point specifically.)

mirdono (commented):
Left two questions I have regarding this migration. Functionally speaking it is fine.

config/migrations/20260407111121-create-impact.sparql
Rahien: i was being hypocritical here and left this spec out so you wouldn't see them in the list of concept schemes, but you're right that it should actually just be there too

config/migrations/20260407111121-create-impact.sparql
Rahien: we decided to have the ai service only label decisions with an sdg if there is impact. If we don't know whether the impact is positive or negative, the impact itself is left blank

mirdono (approved):
Thanks for addressing my questions, and reminding me of decisions I should have remembered (or at least noted down somewhere.)

PR #81, opened 2026-04-09
Add shacl for organizations

brechtvdv:
## 🗒️ Description

This PR adds a new shape to validate the structure of organizations.

## 🦮 How to test

The best way to test the new SHACL shape is by setting up a temporary graph:

1. Insert organization that is good, and one that is bad:
```
prefix sh: <http://www.w3.org/ns/shacl#> 
prefix xsd: <http://www.w3.org/2001/XMLSchema#> 
prefix dct: <http://purl.org/dc/terms/> 
prefix eli: <http://data.europa.eu/eli/ontology#> 
prefix mu: <http://mu.semte.ch/vocabularies/core/> 
prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> 
prefix epvoc: <https://data.europarl.europa.eu/def/epvoc#> 
prefix org: <http://www.w3.org/ns/org#> 
prefix skos: <http://www.w3.org/2004/02/skos/core#> 
prefix dcat: <http://www.w3.org/ns/dcat#> 

INSERT DATA {
GRAPH <http://mu.semte.ch/graphs/temp/org> {
<http://example.org/good-org> a org:Organization ;
 mu:uuid "123" ;
 dct:identifier "123";
 skos:prefLabel "My good org"@en .

<http://example.org/bad-org> a org:Organization ;
 dct:identifier "123";
 skos:prefLabel "My bad org has no uuid" .
}
}
```
2. Adapt the named graphs in `config/reports/shacl-report.js`:
```
const namedGraphs = [
    'http://mu.semte.ch/graphs/temp/org'
];
```
3. Restart the `report-generation` service: `drc restart report-generation`
4. Check the latest report `http://localhost/shacl-reports/shacl-reports/latest/issues`. This should result in two errors:
image

### DIFF
--- a/compose/reporting.yml
+++ b/compose/reporting.yml
@@ -1,6 +1,6 @@
 services:
   report-generation:
-    image: lblod/loket-report-generation-service:0.8.6-beta-6
+    image: lblod/loket-report-generation-service:0.8.6-beta-7
     environment:
       DEFAULT_GRAPH: "http://mu.semte.ch/graphs/reports"
       ONLY_KEEP_LATEST_REPORT: true
--- a/compose/reporting.yml
+++ b/compose/reporting.yml
@@ -1,6 +1,6 @@
 services:
   report-generation:
-    image: lblod/loket-report-generation-service:0.8.6-beta-5
+    image: lblod/loket-report-generation-service:0.8.6-beta-6
     environment:
       DEFAULT_GRAPH: "http://mu.semte.ch/graphs/reports"
       ONLY_KEEP_LATEST_REPORT: true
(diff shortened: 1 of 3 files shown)


Rahien (approved):
This works, but when i run the report generation service on an empty db without changing the graphs, it returns a 500 'something went wrong' without anything useful in the logs. Could it be that when there are no items, the service fails without logs and returns a 500?

brechtvdv:
a fix has been added that returns 404 when report is not found
empty results lead to 200 with empty array
https://github.com/lblod/loket-report-generation-service/pull/16/changes/f24741010c845d5e3f49de8b9349a494037c75b8

PR #84, opened 2026-04-13
Add altLabel and notation to SDG concepts

brechtvdv:
Currently, the “simple” SDG codelist contains the skos:prefLabel (and definition) of each SDG:

[app-decide/config/migrations/add-sdg-codelist/20260310123607-add-simple-sdg-codelist.ttl at development · lblod/app-decide](https://github.com/lblod/app-decide/blob/development/config/migrations/add-sdg-codelist/20260310123607-add-simple-sdg-codelist.ttl) 

However, in the frontend of the impact reporting tool, a shorter label must be shown according to the design:

image

These shorter labels are available in the original SDG codelist with predicate skos:altLabel:

[app-decide/config/migrations/add-sdg-codelist/20260112102218-sdg-codelist.ttl at development · lblod/app-decide](https://github.com/lblod/app-decide/blob/development/config/migrations/add-sdg-codelist/20260112102218-sdg-codelist.ttl) 

Add skos:altLabel to the simple SDG codelist so it can be reused in the frontend.

Also extend the codelist with skos:notation so that each SDG includes its numeric identifier (1–17). This is required for:

Mapping each SDG to its corresponding color in the UI

Ensuring correct ordering of SDGs in the frontend


Andresdev02 (approved):
Looks good to me thanks! 🚀

PR #85, opened 2026-04-14
LBRON-1346 Added smart-search to dispatcher and compose

tdn:
To test, go to http://smart-search.localhost/

This uses v0.0.1 of https://github.com/lblod/frontend-decide-question-answering, which should be included in the review of this first PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,53 @@ There are two main pain points:
 1. Mac has an arm64 processor, a lot of the services don't have a multi-platform image. In the case they only have a amd64 image, docker will gave you a warning about this. In general this is not a real problem since your macbook can just emulate amd64, but still the warnings are annoying, so these are suppressed.
 2. At the moment this project was setup the service mu-identifier weren't working for mac (at least on my device), so you have to build these yourself, and gave them the appropriate image name and tag.
 
+### Running the stack with smart search/question-answering
+
+To get the stack to work properly, including its AI question-answering service, there are a few extra steps that need to be done.
+
+First, add your LLM of choice (e.g., `gemma3:1b`) to your `docker-compose.override.yml`:
+
+```
+question-answering:
+  image: semanticai/decide-question-answering:latest
+  environment:
+    SEARCH_API_URL: "http://search:80/expressions/large-search"
+    EMBEDDING_API_URL: "http://embedding:80/embed"
+    MU_SPARQL_ENDPOINT: "http://database:8890/sparql"
+    MU_SPARQL_TIMEOUT: "30"
+    GENERATION_PROVIDER: "ollama"
+    GENERATION_ENDPOINT: "http://ollama:11434"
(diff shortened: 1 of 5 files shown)


compose/smart-search.yml
```diff
@@ -0,0 +1,6 @@
+services:
```
JonasVanHoof, line 1: Will this compose file grow over time with more smart search things? Or will the search service be moved from the other compose files? (saw the search profile)

compose/smart-search.yml
```diff
@@ -0,0 +1,6 @@
...
+  frontend-smart-search:
```
JonasVanHoof, line 2: Sorry i added this comment in the frontend PR 😁 

Update the dev compose with
```yml
services:
  frontend-smart-search:
    restart: "no"
```

config/dispatcher/dispatcher.ex
```diff
@@ -401,6 +401,19 @@ defmodule Dispatcher do
...
+  # smart search
+  match "/index.html",  %{reverse_host: ["smart-search" | _rest], layer: :static} do
+    forward(conn, [], "http://frontend-smart-search/index.html")
+  end
+
+  get "/assets/*path", %{reverse_host: ["smart-search" | _rest], layer: :static} do
+    forward(conn, path, "http://frontend-smart-search/assets/")
+  end
+
+  get "/@appuniversum/*path",  %{reverse_host: ["smart-search" | _rest], layer: :static} do
```
JonasVanHoof, line 413: Just a question: Why do we need these endpoints? I saw that this is added for every frontend in the stack 🤔

README.md
```diff
@@ -50,6 +50,53 @@ There are two main pain points:
...
+docker compose exec -T ollama ollama pull gemma3:1b
+```
+
+Finally, you need to restart the `embedding` service and ensure it runs error-free:
+```
+docker compose restart embedding
+docker compose logs -f embedding
+```
+
+The frontend for the Smart Search should now be available at http://smart-search.localhost .
```
JonasVanHoof, line 98: This one is not running for me but not sure why yet, could be on me

JonasVanHoof (commented):
Still running into issues for getting an answer from the question-answering-service but here are some comments!

compose/smart-search.yml
tdn: Possibly. In any case, I thought this was sufficiently different to regular (elastic) search to merit its own compose file

config/dispatcher/dispatcher.ex
tdn: So URLs like http://smart-search.localhost work

README.md
tdn: It was running for me and Pedro, so I suspect this to be on your end. Happy to try and resolve it when you're back from leave

Rahien (approved):

PR #88, opened 2026-04-23
feat: add annotation-job-splitter service

mirdono:
Adds the new [annotation-job-splitter](https://github.com/lblod/annotation-job-splitter-service) service to this application.

This service will be responsible to split annotation jobs into appropriate subtasks that can be executed by services in the (AI) pipelines. To this end, the configuration for the `deltanotifier` has been updated to forward the necessary delta messages to the splitter service.

## Notes
The updated job-controller configuration relies on the features added in [#9](https://github.com/lblod/job-controller-service/pull/9)

## How to test
See the PR on the service's initial implementation: [#1](https://github.com/lblod/annotation-job-splitter-service/pull/1)

## TODO
- [x] Use a tagged version of the new service instead of a feature build

## Related tickets
- LBRON-1311
- LBRON-1373

### DIFF
--- a/compose/pipeline.yml
+++ b/compose/pipeline.yml
@@ -36,7 +36,7 @@ services:
     labels:
       - 'logging=true'
   job-controller:
-    image: lblod/job-controller-service:1.2.0
+    image: lblod/job-controller-service:1.3.0-beta-parallel-tasks
     volumes:
       - ../config/job-controller:/config
     environment:
--- a/compose/pipeline.yml
+++ b/compose/pipeline.yml
@@ -56,3 +56,10 @@ services:
       - 'logging=true'
     restart: always
     logging: *default-logging
+  annotation-job-splitter:
+    # TODO: Change to tagged version once available
+    image: lblod/annotation-job-splitter-service:feature-initial-implementation-tasks
+    restart: always
+    logging: *default-logging
+    labels:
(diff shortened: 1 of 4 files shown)


config/delta/rules.js
```diff
@@ -12,4 +12,5 @@ export default [
   ...pdfToEli,
   ...osloToEli,
   ...search,
+  ...annotationJobSplitter,
```
Rahien, line 16: import still missing

mirdono: Huh, no idea how I lost this import. Must have had it before committing as it did work as expected the whole time while developing the service.

compose/pipeline.yml
```diff
@@ -56,3 +56,10 @@ services:
       - 'logging=true'
     restart: always
     logging: *default-logging
+  annotation-job-splitter:
+    # TODO: Change to tagged version once available
```
Rahien: my bad but when changing this, could you also set the job-controller to [1.3.0-beta-parallel-tasks](https://hub.docker.com/layers/lblod/job-controller-service/1.3.0-beta-parallel-tasks/images/sha256-7650187f860effebc292ec29bc8f7ad509107d274e4d0d2d056f903444565aeb)

Rahien (approved):

PR #94, opened 2026-05-08
add splitter config

Rahien:


config/annotation-job-splitter/config.ts
```diff
@@ -0,0 +1,40 @@
...
+        ],
+      },
+    'http://lblod.data.gift/id/jobs/concept/JobOperation/codelist-matching/evaluation':
+      {
+        taskConfiguration: [
+          {
+            currentOperation:
+              'http://lblod.data.gift/id/jobs/concept/TaskOperation/codelist-matching/evaluation-split-tasks',
+            nextOperation:
+              'http://lblod.data.gift/id/jobs/concept/TaskOperation/codelist-matching/annotate',
```
mirdono, line 21: **Issue:** The job-controller expects `CLASSIFIER-annotate` as task operation.

The job-controller is [configured](https://github.com/lblod/app-decide/blob/karel/add-splitter-config/config/job-controller/config.json#L176-L177) with a different task operation here. This mismatch will mess up running the pipeline.

config/annotation-job-splitter/config.ts
```diff
@@ -0,0 +1,40 @@
...
+          {
+            currentOperation:
+              'http://lblod.data.gift/id/jobs/concept/TaskOperation/annotation-split-tasks',
+            nextOperation:
+              'http://lblod.data.gift/id/jobs/concept/TaskOperation/translating',
+          },
+        ],
+      },
+  },
+  // optional
```
mirdono: **Suggestion:** Remove these commented lines.

I left these lines in the service's default/example configuration for illustration purposes.  Since this is an app-specific configuration I would remove them here, as to avoid any confusion on whether we are using/have always used the default values for these predicates.  If it is necessary to change them in the future, the service README describes how.

mirdono (changes requested):
Added inline comments about an issue (aside from the missing PR description :wink: ) and a suggestion.

config/annotation-job-splitter/config.ts
Rahien: i added the job-controller changes from test to this pr as well

mirdono (approved):

PR #97, opened 2026-05-08
LBRON-1364 smart search annotation datamodel

tdn:
*** Do not merge yet***
Due to the extensions needed for this use case, we need to make a few changes to the HVT frontend and annotation-review service to make sure everything keeps working the way it does now.

To align with the way we store annotations (and reviews) in the HVT (see https://github.com/lblod/frontend-decide-human-validator and https://github.com/lblod/annotation-review-service), we choose to use the Web Annotation Ontology: http://www.w3.org/ns/oa#
In the HVT, oa:Annotation is used in 2 forms: 
1. to store the annotations the AI pipeline makes, as oa:Annotation with a nif:confidence score.
2. to store the reviews of the annotations, also as oa:Annotation , with the added class of ext:ReviewAnnotation

Each annotation has a target, which in the HVT was limited to resources of class oa:SpecificResource. To accommodate other annotations/reviews though, we need to allow other resource types as target as well. Therefore, we introduce a parent class ext:AnnotationTarget,  whch oa:SpecificResource can extend. 

For the smart search (UC2), we need to store two types of user feedback:
1. Feedback (thumbs up/down) on the answer as a whole.
2. Feedback (thumbs up/down) on the decisions listed as sources for the answer.

To do this, we introduce two new classes to the model that extend ext:AnnotationTarget : schema:Answer and schema:Quotation . 

The model for annotations is summarized in the following diagram:
image

notes:
- possibly, the LLM should be referenced through prov:wasAssociatedWith ?agent as in https://github.com/lblod/annotation-review-service/blob/master/controllers/annotations.ts#L357 . However, we need to confirm that the question-answering service has the correct agent URI

- We need to add a few small changes and a thorough test of frontend-decide-human-validator and annotation-review to accomodate these extensions, and to ensure the HVT keeps working the way it does.
PR: https://github.com/lblod/frontend-decide-human-validator/pull/18


brechtvdv:
Can you add the review result (http://mu.semte.ch/vocabularies/ext/annotation-review#approve and #reject) as subclasses of AnnotationBody?
Human validation tool still works with these changes

Rahien (changes requested):
It seems that with the added superclass, the annotation-review service now returns duplicate results in the response when listing the annotations for an expression. could you have a look at that?

Rahien (approved):
can't reproduce the issue i had with duplicate annotations

tdn:
> Can you add the review result (http://mu.semte.ch/vocabularies/ext/annotation-review#approve and #reject) as subclasses of AnnotationBody? 

you mean just in the diagram? Or also in the resource config? Because the result is just a constant URI, right?

brechtvdv:
Fyi: the download script is using this datamodel for human validations: https://github.com/lblod/app-decide/pull/105

brechtvdv (approved):

PR #55, opened 2026-02-14
Adding NER service and dependencies of NER/NEL

javdrher:
This PR introduces the NER service, and adds some dependencies of the AI team

- Nominatim
- QDrant

It does not fully cover additional needs from ollama for the NEL service

PR #43, opened 2026-02-03
add mu search and embeddings embeddings

Rahien:
Adds mu-search and embeddings service to the app (disabled by default in docker-compose.dev.yml)

Generating embeddings will take a long time, especially when running without GPU. I'll be moving my embeddings generated on GPU to the dev server manually.

PR #24, opened 2025-12-02
feat: add support DSP Catalog protocol

mirdono:
Extend the application to support the DSP Catalog protocol implemented by the DCAT-service in [#2](https://github.com/lblod/dcat-service/pull/2). This PR adds a route for the service's new functionality, and configures the appropriate environment variables for the service. Note that the set `BASE_URL` is an example, and might change in the future.

Furthermore, this PR adds some migrations with additional example data to better illustrate the service's DSP Catalog protocol functionality. This data is for example/testing purposes only, and should **not** be merged into the application.

## Notes
This PR also introduces two small unrelated changes:
- Silence warnings by the dispatcher about unused `path` variables by prefixing them with a `_`. This makes the dispatcher output a little bit cleaner.
- In a DEV setup, expose port 8890 of the `virtuoso` service. This makes it a  it easier to directly query the triplestore during development.

## Related PRs
- [DCAT service](https://github.com/lblod/dcat-service/pull/2) extension for DSP Catalog Protocol

## Related tickets
- LBRON-784

PR #101, opened 2026-05-18
chore: document application setup (for partners)

mirdono:
This PR extends the existing documentation to better support partners to set up their own instance of the DECIDe application.  This initially focuses on the city of Bamberg who are primarily interested in use cases 0.1 (policy impact report) and 2 (smart search).  Further guidelines tailored to other partner cities will be added in follow-up PRs.

## Proposed approach
This PR mainly introduces two new files in the new `docs` folder.  The [README](https://github.com/lblod/app-decide/blob/chore/document-app-setup/docs/README.md) provides additional information concerning setting up an own, non-development, instance of the DECIDe application.

Furthermore, a Bamberg-specific [docker compose override](https://github.com/lblod/app-decide/blob/chore/document-app-setup/docs/docker-compose.override.bamberg.yml) configuration was added. This configuration (partially) disables services that are not relevant for Bamberg and contains some template configuration for relevant services.

Furthermore, the primary [README](https://github.com/lblod/app-decide/blob/chore/document-app-setup/README.md) was updated with respect to
- add links to used services where possible;
- improved the documentation PDF harvesting pipeline (UC0.0);
- slightly extended the documentation for UC0.1 to included the used services; and
- documented UC2 smart search and revised the explanation to set up the smart search in a development stack.

## How to test
- Read all added documentation and assess whether it is factually correct and complete.
- Add the Bamberg-specific docker compose override to your `.env` and ensure the stack can be properly started.  Note, it is expected that some services do not function correctly due to placeholder values in their configuration.

## Notes
In addition to the changes above, this PR also slightly simplifies the docker-compose configuration:
- Merged `smart-search.yml` into `search.yml` as the `frontend-smart-search` configured in the former relates to the services configured in the latter.
- Moved the configuration for the `report-generation` service into `base-compose.yml` to reduce the number of configuration files.

## Related tickets
- LBRON-1457


README.md
```diff
@@ -137,21 +183,41 @@ To start the OParl pipeline, create a "Harvest OParl API & Publish as ELI" job i
...
+
+For a more detailed description of this use case and how it was developed withing the DECIDe project see the dedicated [gitbook page](https://app.gitbook.com/o/-MP9Yduzf5xu7wIebqPG/s/PzeOtGh2pfnNKyqa7G5w/decide-project/write-up-uc2-smart-search).
+
+For more information on configuring an app instance to support smart search see the [above configuration section](#running-the-stack-with-smart-searchquestion-answering).
+>>>>>>> 02ccdbd (chore(doc): revise section on smart search setup)
```
mirdono: **Issue:** Lines should be removed.

These line are probably a leftover cleaning up the history and should be removed.

mirdono (commented):
Added some issues resulting from feedback via rocket.chat. (Along with one that caught my eye while adding these comments.)

docs/README.md
```diff
@@ -0,0 +1,167 @@
...
+|---------------------------------------------------------------------------------------------------------------|------------------------|
+| [Pipeline dashboard](https://github.com/lblod/frontend-harvesting-self-service/tree/feature/oparl-harvesting) | `dashboard`            |
+| [Yasgui](https://github.com/lblod/frontend-decide-yasgui)                                                     | `yasgui`               |
+| [dcat](https://github.com/lblod/frontend-decide-dcat)                                                         | `ds`                   |
+| [Human Validation Tool](https://github.com/lblod/frontend-decide-human-validator)                             | `human-validator`      |
+| [Smart search](https://github.com/lblod/frontend-decide-question-answering)                                   | 'smart-search'         |
+| [Policy impact report](https://github.com/lblod/frontend-decide-policy-impact-report)                         | `policy-impact-report` |
+
+
+### Outsource LLM to the cloud
```
mirdono, line 76: **Issue:** Clarify the need to obtain API keys for external services and how they are to be configured.

As part of [LBRON-1423](https://binnenland.atlassian.net/browse/LBRON-1423) the API keys will become environment variables to be configured in a non-versioned override file.  This would be similar as the `entity-linking` configuration in the added example override.

Furthermore, it is probably best to emphasise these API keys are configured into a local `docker-compose.override.yml` instead of the versioned example override.

docs/README.md
```diff
@@ -0,0 +1,167 @@
...
+- `git` to obtain the application source code
+- `docker` and `docker compose` to configure and run the application's microservices
+- A reverse proxy that forwards HTTP requests to the app's identifier service. We typically use [app-letsencrypt](https://github.com/redpencilio/app-letsencrypt) for this purpose.
+
+### Updating the app
+Generally updating (parts of) the app consists of pulling the latest version from the remote repository via a  `git pull` and, recreating and/or restarting the appropriate services.
+For each service `A` that was added or updated (version bump or changed environment variables), do `docker compose up [-d] A`. For each service `B` for which their configuration was updated in the `../config/B` folder, do a `docker compose restart B`. Note, that `up` on its own does **not** cause a service to update its configuration.
+
+
+## Service configuration
```
mirdono, line 30: **Issue:** Add a section for the `ollama` service.

The `ollama` service should pull in the appropriate LLMs, as described in [#55](https://github.com/lblod/app-decide/pull/55#issuecomment-4328654240).  While the general README mentions this for the smart search setup it is currently not documented this is needed for other use cases as well.

mirdono:
**Issue:** Document `municipality-linker` service.

In [#99](https://github.com/lblod/app-decide/pull/99) a `municipality-linker` service was added.  Check whether this requires further configuration and/or explanation in the context of this PR.

mirdono:
**Remark:** Add documentation/pointer concerning task/job data model.

The pipelines with their jobs and tasks are central to the application but are not yet explicitly explained here.  This came up in the project's get-together today.

docs/README.md
Rahien: i flagged this with pedro and joachim, the service should have ollama load the models on startup if they aren't available yet in ollama on disk and crash if that fails. Running manual pulls by drc execing seems crazy

Rahien, line 30: maybe add a reference here to the partner configurations and their docker-compose.overrides?

docs/README.md
```diff
@@ -0,0 +1,167 @@
...
+
+Finally, restart the `migrations` service to execute the created migration.
+
+```bash
+docker compose restart migrations; docker compose logs -f migrations
+```
+
+
+### Verifiable credentials
+- TODO VC configuration
```
Rahien: i don't think any partner will actually run this?

mirdono: I do not think so. But since it is part of the dataspace side of the project documentation on how one could set it up seems a nice to have imo. At least given it is doable, I seem to recall it involved custom builds of apps, so maybe we it would bring us too far wrt the goal of this documentation.

docs/README.md
mirdono: Updated the example override entry for `named-entity-recognition` based on [#43](https://github.com/semantic-ai/decide-geocoding-service/pull/43)

mirdono:
> **Issue:** Document `municipality-linker` service.
> 
> In [#99](https://github.com/lblod/app-decide/pull/99) a `municipality-linker` service was added.  Check whether this requires further configuration and/or explanation in the context of this PR.

This service works as is, no partner-specific configuration needed.

mirdono:
> **Remark:** Add documentation/pointer concerning task/job data model.
> 
> The pipelines with their jobs and tasks are central to the application but are not yet explicitly explained here.  This came up in the project's get-together today.

Added a short background section in the UC0.0 section. This very shortly introduces the essentials of the pipelines and provides a link to the more elaborate pipeline write up.

docs/README.md
mirdono: Gave it some more thought and I removed the VC section placeholder for now.  Should we need documentation on this later on we can always add it.

docs/README.md
mirdono: Updated the example override entry for `pdf-content` based on [#24](https://github.com/semantic-ai/decide-pdf-content-extraction/pull/24)

mirdono:
As discussed on standup of 28 May 2026, merged without explicit review as we want to get it out to the partners.

PR #103, opened 2026-05-19
add example migration for the non-public sparql endpoint

Rahien:
this adds a migration that describes the (not yet implemented) private sparql endpoint and adds an ODRL policy to it that describes that no one can use it except for members of the licensedUserCollection, which is an odrl:PartyCollection

Non-public datasets can easily be discovered by checking if they are the target of an odrl:Policy that has an odrl:prohibition without odrl:assignee (so default anyone), that forbids the odrl:use action.

### DIFF
--- a/config/migrations/20260519091850-add-private-dataset.ttl
+++ b/config/migrations/20260519091850-add-private-dataset.ttl
@@ -1,0 +1,75 @@
+@prefix odrl: <http://www.w3.org/ns/odrl/2/> .  
+@prefix dcat: <http://www.w3.org/ns/dcat#> .  
+@prefix dct: <http://purl.org/dc/terms/> .  
+@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .  
+@prefix foaf: <http://xmlns.com/foaf/0.1/> .  
+@prefix schema: <https://schema.org/> .  
+@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
+@prefix private-ds-ex: <http://decide.data.gift/datasets/example/60e19fdf-b549-41b9-9adf-420379285127/> .
+@prefix public-ds-ex: <http://decide.data.gift/datasets/example/1ae2da8e-9889-426b-a7e8-9360c15ba2cd/> .
+@prefix mu: <http://mu.semte.ch/vocabularies/core/> .
+@prefix ext: <http://mu.semte.ch/vocabularies/ext/> .
+
+public-ds-ex:dataset a dcat:Dataset ;  
+    mu:uuid "1ae2da8e-9889-426b-a7e8-9360c15ba2cd" ;
+    dct:title "Licensed Dataset" ;  
+    dct:description "An example dataset available only to users who purchased a valid license." ;  
+    dcat:distribution public-ds-ex:distribution .
+
+public-ds-ex:distribution a dcat:Distribution ;  
+    mu:uuid "12f78fc1-9292-48fd-8c91-e835578dfcf3" ;
+    dct:format "application/sparql-query" ;  
+    dcat:accessURL <https://ds.decide.lblod.info/api/sparql> .
+
+public-ds-ex:policy a odrl:Agreement, ext:PublicPolicy ;  
+    mu:uuid "ab4509a9-4676-4966-bc07-8c20a3eb1d80" ;
+    dct:description "Anyone may use this dataset" ;
+    odrl:target public-ds-ex:dataset ;  
+    odrl:assigner <http://ds.decide.lblod.info> ;  
+    odrl:permission [  
+        odrl:action odrl:use ;  
+    ].
+
+
+private-ds-ex:dataset a dcat:Dataset ;  
+    mu:uuid "60e19fdf-b549-41b9-9adf-420379285127" ;
+    dct:title "Licensed Dataset" ;  
+    dct:description "An example dataset available only to users who purchased a valid license." ;  
+    dcat:distribution private-ds-ex:distribution .
+
+private-ds-ex:distribution a dcat:Distribution ;  
+    mu:uuid "341bc1ef-f4ab-4450-80af-b72d7c682e7f" ;
+    dct:format "application/sparql-query" ;  
+    dcat:accessURL <https://ds.decide.lblod.info/api/private/sparql> .
+
+private-ds-ex:policy a odrl:Agreement, ext:RestrictedPolicy ;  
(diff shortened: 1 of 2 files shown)


config/migrations/20260519091850-add-private-dataset.ttl
```diff
@@ -0,0 +1,75 @@
...
+
+public-ds-ex:distribution a dcat:Distribution ;  
+    mu:uuid "12f78fc1-9292-48fd-8c91-e835578dfcf3" ;
+    dct:format "application/sparql-query" ;  
+    dcat:accessURL <https://ds.decide.lblod.info/api/sparql> .
+
+public-ds-ex:policy a odrl:Agreement, ext:PublicPolicy ;  
+    mu:uuid "ab4509a9-4676-4966-bc07-8c20a3eb1d80" ;
+    dct:description "Anyone may use this dataset" ;
+    odrl:target public-ds-ex:dataset ;
```
brechtvdv: I think odrl:target and odrl:assigner should be placed inside odrl:permission
(https://www.w3.org/TR/odrl-model/)

Rahien: I had read that 
> An ODRL Policy MAY also declare properties which are shared and common to all its Rules. Specifically; action properties, sub-properties of relation (such as target), and sub-properties of function (such as assigner and assignee). See section [Compact Policy](https://www.w3.org/TR/odrl-model/#composition-compact) for validation requirements on these shared properties.

And that did feel more clear to me then repeating them in the permission/prohibition, no?

config/migrations/20260519091850-add-private-dataset.ttl
```diff
@@ -0,0 +1,75 @@
...
+    dct:format "application/sparql-query" ;  
+    dcat:accessURL <https://ds.decide.lblod.info/api/sparql> .
+
+public-ds-ex:policy a odrl:Agreement, ext:PublicPolicy ;  
+    mu:uuid "ab4509a9-4676-4966-bc07-8c20a3eb1d80" ;
+    dct:description "Anyone may use this dataset" ;
+    odrl:target public-ds-ex:dataset ;  
+    odrl:assigner <http://ds.decide.lblod.info> ;  
+    odrl:permission [  
+        odrl:action odrl:use ;
```
brechtvdv: I would use odrl:read, as they probably can't insert or delete data

config/migrations/20260519091850-add-private-dataset.ttl
```diff
@@ -0,0 +1,75 @@
...
+    mu:uuid "341bc1ef-f4ab-4450-80af-b72d7c682e7f" ;
+    dct:format "application/sparql-query" ;  
+    dcat:accessURL <https://ds.decide.lblod.info/api/private/sparql> .
+
+private-ds-ex:policy a odrl:Agreement, ext:RestrictedPolicy ;  
+    mu:uuid "1d8e1e34-ac8c-4fb8-845a-a9e8ca0e9f6b" ;
+    dct: "Only licensed users may access this content" ;
+    odrl:conflict odrl:perm ;
+    odrl:assigner <http://ds.decide.lblod.info> ;  
+    odrl:target private-ds-ex:dataset ;
```
brechtvdv: move target inside odrl:permission (correct me if I'm wrong)

config/migrations/20260519091850-add-private-dataset.ttl
Rahien: also specified that write is not allowed then

config/migrations/20260519091850-add-private-dataset.ttl
```diff
@@ -0,0 +1,75 @@
...
+    dct:description "Collection of users or organizations who have purchased a valid license for the dataset." ;  
+    dct:relation private-ds-ex:licenseOffering . 
+
+# How to obtain the license, url still to be determined
+private-ds-ex:licenseOffering a schema:Offer ;  
+    mu:uuid "982416d7-5702-41fe-a08a-9f28542e5c82" ;
+    schema:name "Dataset Access License" ;  
+    schema:description "A license granting access to the licensed dataset." ;  
+    schema:url <http://ds.decide.lblod.info/licenses/purchase> ;
+    schema:howTo [
```
brechtvdv: schema:howTo does not exists

Maybe use schema:about on the HowTo?

config/migrations/20260519091850-add-private-dataset.ttl
brechtvdv: I would follow the approach (repeating) of other data space stuff, for example:
- Fiware: https://github.com/SEAMWARE/odrl-pap#try-out-some-policies
- Gaia-x: https://gitlab.com/gaia-x/lab/policy-reasoning/odrl-vc-profile/-/tree/main/examples
- Idlab: https://spec.knows.idlab.ugent.be/force/all/89e098e0fb30b31943b604407af44dfe88a3741e/

config/migrations/20260519091850-add-private-dataset.ttl
```diff
@@ -0,0 +1,78 @@
...
+    dct:description "Collection of users or organizations who have purchased a valid license for the dataset." ;  
+    dct:relation private-ds-ex:licenseOffering . 
+
+# How to obtain the license, url still to be determined
+private-ds-ex:licenseOffering a schema:Offer ;  
+    mu:uuid "982416d7-5702-41fe-a08a-9f28542e5c82" ;
+    schema:name "Dataset Access License" ;  
+    schema:description "A license granting access to the licensed dataset." ;  
+    schema:url <http://ds.decide.lblod.info/licenses/purchase> ;
+    schema:about [
```
brechtvdv: I meant the other direction.

schema:HowTo - schema:about -> schema:Offer

config/migrations/20260519091850-add-private-dataset.ttl
```diff
@@ -0,0 +1,78 @@
...
+    odrl:permission [  
+        odrl:action odrl:read ;  
+        odrl:assignee private-ds-ex:licensedUserCollection ;  
+    ] ;
+    odrl:prohibition [  
+        odrl:action odrl:read, odrl:write ;  
+    ] .
+
+# PartyCollection listing the users with a license
+private-ds-ex:licensedUserCollection a odrl:PartyCollection ;
```
brechtvdv, line 67: Do these users require a certain VC?
We could express this with the VCO: https://gitlab.com/gaia-x/lab/policy-reasoning/odrl-verifiable-credential-ontology-mapping

Rahien: no, it's up to the implementation by UI!, but yes in theory it could be done with a VC.

brechtvdv (approved):

PR #107, opened 2026-05-26
add dsp-auth-wrapper in preparation for the dsp connector

Rahien:
This add some example content to the restricted graph `<http://mu.semte.ch/graphs/organizations/353234a365664e581db5c2f7cc07add2534b47b8e1ab87c821fc6e6365e6bef5/Decide-gebruiker>` that people can get access to through the VC or the [dsp-auth-wrapper service](https://github.com/lblod/decide-dsp-authorization-wrapper).

It also adds a scope for that service and grants read access to public, the organizations graph and read and write access to the session graph for that service.

When activating this service, one should be able to follow the setup in the dsp-auth-wrapper service to configure a proper api key and then send a sparql query using `POST http://localhost/api/private/sparql`.  When asking for the example data in the private graph, the public sparql endpoint should return an empty resultset, while the private sparql endpoint should return the correct triples.

### DIFF
--- a/compose/data-space.yml
+++ b/compose/data-space.yml
@@ -68,3 +68,11 @@ services:
       - 'logging=true'
     restart: always
     logging: *default-logging
+  dsp-auth-wrapper:
+    image: lblod/decide-dsp-authorization-wrapper:0.0.1
+    labels:
+      - 'logging=true'
+    restart: always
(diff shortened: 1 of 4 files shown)


mirdono (approved):
Tested and works as expected (after some messing around to send a correct request)

PR #112, opened 2026-05-27
Update section on OSLO pipeline in README

MartijnBogaert:
The section on the OSLO-ELI pipeline in the README was outdated. This PR brings it up to date with how the corresponding services currently work. E.g. it is no longer necessary to manually set env variables in the override to switch between initial and delta sync modes: this choice can now easily be made from the harvesting frontend.


mirdono (approved):
Thanks for updating this!

PR #115, opened 2026-05-28
Fixes to download dataset

brechtvdv:
- Work MUST exists for every expression, so not making this optional
- Uses eli:passed_by for finding the municipality (ext:owningBody is only when entity linking is used)
- Generalizes the codelist query (and adds restricted mobility zones)
- Extends human validation for smart search and makes sure body is also exported


JonasVanHoof (approved):
Very nice additions to the queries. Yesterday we went over them together and tweaked them a bit. Cleaned up the code this morning (rebase, small readme update) and all still runs and outputs the expected (expect for rmz, HV as we do not have data for this)

Good job, thanks! 💪🏻

PR #118, opened 2026-06-02
Fix/brecht/lbron 1561 fix shacl validation

brechtvdv:
Fixes bug in shacl validation service where a report has multiple uuid's and created timestamps: https://binnenland.atlassian.net/browse/LBRON-1561

One timestamp is now added after the validation round
The UUID of the report is passed to the validation for reuse

Depends on: https://github.com/lblod/loket-report-generation-service/pull/17

Also, the title property of a eli:Work is expected at least once (before it was twice)

### DIFF
--- a/compose/reporting.yml
+++ b/compose/reporting.yml
@@ -5,7 +5,7 @@ x-logging: &default-logging
     max-file: '3'
 services:
   report-generation:
-    image: lblod/loket-report-generation-service:0.8.6-beta-8
+    image: lblod/loket-report-generation-service:0.8.7
     environment:
       DEFAULT_GRAPH: 'http://mu.semte.ch/graphs/reports'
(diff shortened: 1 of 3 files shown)


compose/reporting.yml
```diff
@@ -5,7 +5,7 @@ x-logging: &default-logging
     max-file: '3'
 services:
   report-generation:
-    image: lblod/loket-report-generation-service:0.8.6-beta-7
+    image: lblod/loket-report-generation-service:0.8.6-beta-8
```
Rahien: should we still use betas? why not bump after merging the most recent pr?

Rahien (approved):

compose/reporting.yml
brechtvdv: You're right, now it's time to see goodbye to those betas

PR #32, opened 2025-12-19
feat: add annotations and geo data

mirdono:
This builds on [#27](https://github.com/lblod/app-decide/pull/27) to provide a public endpoint for the app's annotations with geographical information. More specifically, this extends the configuration for the resource and dispatcher services with the appropriate resource definitions and routes respectively.  Furthermore, this bumps the used `docker-virtuoso` to a version supporting geosparql and updates its configurations to load the appropriate plugins. Finally, this adds a migration inserting some example data for demonstration and testing purposes.

## How to test
This PR involves upgrading the version of the virtuoso service. For this follow the [upgrade instructions](https://github.com/redpencilio/docker-virtuoso?tab=readme-ov-file#upgrading) in the `docker-virtuoso` README. Note, make sure you already started the virtuoso service **before** switching to this PR's branch.

Afterwards, make sure to restart the `resources` and `dispatcher` services such that they load their updated configurations. To load the example data into the app simply start the migrations service.

## Notes
- The data model is documented on [gitbook](https://app.gitbook.com/o/-MP9Yduzf5xu7wIebqPG/s/grRYjcouX0uZY2P6PYBl/business-analysis-research/decide-use-cases/uc1-mapping-ld-and-l-on-restricted-mobility-zones-to-geo-locations/review-prototype-of-ai-team-to-plot-decisions-to-geolocation)

## Related PRs
- Frontend PR: [#8](https://github.com/lblod/frontend-decide/pull/8)

## Related tickets
- LBRON-920

PR #126, opened 2026-06-08
remove example data

Rahien:
Removes the example data we added in these prs:
https://github.com/lblod/app-decide/pull/12
https://github.com/lblod/app-decide/pull/32


mirdono (approved):
A much needed cleanup, thanks for taking care of this.

PR #135, opened 2026-06-15
fix: proper account management for pipeline dashboard

mirdono:
The app contained a default `test` account for the pipeline (harvester) dashboard.  This silently initiates every app instance with such an account.  Furthermore, the versioned migration contained the password hash, thereby effectively leaking it.

This corrects that situation by

- removing the migration adding the `test` account; and
- providing a mechanism to create user accounts for the dashboard.

## Proposed approach
To create accounts we leverage the `generate-account` script provided by the [registration](https://github.com/mu-semtech/registration-service), this already does the heavy lifting of generating a correct migration file.  To this end the registration service was added to the `docker-compose.dev.yml` config so this automatically available in development environments.  To remove accounts we provide an example query which can be copied into a migration and only requires the necessary account UUID(s) to be added.  The README is extended with explanations of how to go about adding and removing accounts

Furthermore, a `local` folder is added to `config/migrations` in which migrations can be placed that should not be versioned.  For example, because they contain sensitive data such as password hashes or should be executed only on specific environments.

Finally, this PR removes the migration that introduced the `test` account to prevent future app instances are still initialised with this account.  Deleting the `test` account for existing app instances will be done in a [follow-up PR](https://github.com/lblod/app-decide/pull/136).  This allows to provide some margin for people to create their own accounts before it is removed.

## How to test
Go through the steps to register and disable accounts as added in the README.  Check that they are correct and clear, including for less experienced users.

## TODO
- [ ] Document optional usage of pepper (application salt)

## Related tickets
- LBRON-1589


mirdono (commented):
Added a TODO concerning pepper configuration, following a remark from @Rahien .

README.md
```diff
@@ -53,6 +53,54 @@ This should be your go-to way of starting the stack.
...
+2. Move the generated migration to the `config/migrations/local/` folder of the app instance(s) on which this account should be available.
+3. **Target**: On each target app instance, restart the `migrations` service to execute the generated migrations add inserting the account data: `docker compose restart migrations`
+4. **Target**:  Login into pipeline dashboard of the target app instance using the `USERNAME` and `PASSWORD` specified in step 1 to ensure it works.
+
+#### Disabling an account
+To disable an account its status can be changed to inactive via a migration.
+
+1. Generate a new migration file using the `new` script provided by the `migrations` service: `mu script migrations new sparql NAME`. This will create a blank file `config/migrations/TIMESTAMP-NAME.sparql`, where `TIMESTAMP` is the time you executed the script.
+2. Insert the example query below into the created file and replace `ACCOUNT_UUID` with the UUID for the account to be disabled.
```
brechtvdv, line 73: I think we should explain how the accountUuid can be retrieved (on a server), for example, by running 
`docker exec -it virtuoso isql-v` 
and running following command:

`sparql PREFIX foaf: <http://xmlns.com/foaf/0.1/>
PREFIX mu:      <http://mu.semte.ch/vocabularies/core/>

select ?accountName ?accountUuid
where {
?s a foaf:OnlineAccount ;
foaf:accountName ?accountName ;
mu:uuid ?accountUuid .
};`

mirdono: I added a paragraph explaining where/how to find appropriate UUIDs.  I somewhat modified your example to make it more easily copy-pastable.

Rahien (approved):

PR #145, opened 2026-06-19
have ai services use configured agent uri and custom start.sh script, explain about it in the readme

Rahien:
## description
have ai services use configured agent uri and custom start.sh script, explain about it in the readme

## how to test
Run the start.sh script as explained in the readme, keep the default env vars the same. 
Run a pipeline, notice that the annotations are now generated by a task that has prov:wasAssociatedWith the uri that represents the versioned service with its configuration.

### DIFF
--- a/compose/ai.yml
+++ b/compose/ai.yml
@@ -9,7 +9,7 @@ services:
   # AI PIPELINES
   ##############################################################################
   entity-linking:
-    image: semanticai/entity-linking-backend:0.0.4
+    image: semanticai/entity-linking-backend:0.1.0
     environment:
       TARGET_GRAPH: 'http://mu.semte.ch/graphs/public'
       DEFAULT_GRAPH: 'http://mu.semte.ch/graphs/harvesting'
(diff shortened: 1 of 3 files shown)


compose/ai.yml
```diff
@@ -23,6 +23,9 @@ services:
       LLM_PROVIDER: 'mistral'
       LLM_MAX_RETRIES: 3
       MISTRAL_MODEL: 'ministral-14b-2512'
+      FORCE_VERSIONED_AGENT_URI: true
+      CONFIGURED_AGENT_URI: http://lblod.data.gift/id/components/named-entity-linking/decide${APP_VERSION}
+      CONFIG_REPO_URL: https://github.com/lblod/app-decide${APP_VERSION}
```
mirdono, line 28: **Remark:** Document these environment variables.

These new environment variables are not documented in the individual services, or in this repo.  I think we should at least document somewhere what they mean and when/if their value can be changed.  If they should never be changed, than that should be documented.

mirdono (approved):

PR #146, opened 2026-06-19
fix missing read rights for question answering

Rahien:
question answering isn't allowed to read from public graphs anymore, this fixes that

try by:
running question answering and asking what to take into account to throw a halloween party, you should receive titles and contents in the response

note: your LLM request will likely time out the identifier service, but you can check the logs for that database and see that the query to fetch the ?content now is allowed to read the right graphs


mirdono (approved):
Could not fully test it due to local issues, but changes look correct.

PR #147, opened 2026-06-19
remove decide.lisp in favor of the ttl version

Rahien:
## description
removes the decide.lisp version of the mu-auth config, people sometimes feel they have to keep it in sync with the ttl version, but it already has diverged

## how to test

kill and restart your database after applying this change, make sure you can still access the dashboard and start tasks


mirdono (commented):
When merging [#144](https://github.com/lblod/app-decide/pull/144) I first rebased its branch onto development.  I had not seen you started from that PR for this one.  To fix the resulting conflict, I changed the target branch here to development and rebased the source branch onto development.

mirdono (approved):
I also removed the, now unnecessary, instructions on how to use the lisp config instead of the ttl.  Otherwise it is fine.

PR #151, opened 2026-06-25
fix incorrect predicate in municipality-linker config

Rahien:


Rahien:
closed in favor of https://github.com/lblod/app-decide/pull/152

PR #155, opened 2026-06-30
LBRON-1363 UC2 answer and sources review

tdn:
Adds the `annotation-target` supertype to `annotation`.

Before merging, the HVT needs to be checked so it doesn't store duplicates. However, this should have been addressed by making the relations `polymorphic` in https://github.com/lblod/frontend-decide-human-validator/pull/18

Related PRs:

- https://github.com/semantic-ai/decide-question-answering/pull/14
- https://github.com/lblod/frontend-decide-question-answering/pull/4
- https://github.com/lblod/annotation-review-service/pull/18


Rahien (changes requested):
It seems the review annotations are not actually stored by the service. I.e. I ask for things related to the answer and find nothing after sending an approval

tdn:
> It seems the review annotations are not actually stored by the service. I.e. I ask for things related to the answer and find nothing after sending an approval

Huh. This seems to be the case indeed. But no errors are thrown, so I guess this must be a sparql-parser/odrl thing?

Weird that your HVT annotations are saved, but these aren't.

Rahien (approved):

PR #160, opened 2026-07-06
allow the annotation-review service to make corrections

Rahien:
for description and how to test, see https://github.com/lblod/annotation-review-service/pull/19


tdn (approved):
tested locally and works as expected

PR #162, opened 2026-07-13
ignore normatieve bepaling in hvt

Rahien:
This filters out the normatieve bepaling entities in HVT:

before: 
image

after: 
image


brechtvdv:
We also need to filter out dct:extend, this is a property of the Normative provision

brechtvdv (approved):

PR #164, opened 2026-07-14
fix agent uri filtering in HVT

Rahien:
the agent filter was too restrictive, with the new versioned uris it doesn't hold up. 
This fixes the filter by filtering out agents based on what they are a specialization of

Test it by opening the hvt on a recent dataset and visiting http://human-validator.localhost/validate/87427a74-2bd8-47ba-93ea-7f138e22b22a


brechtvdv (approved):
Looks good!

PR #171, opened 2026-07-24
remove all uses of odrl:write in favor of odrl:modify

Rahien:
For documentation purposes i also modified the (old) migration and changed odrl:write to modify there.

I still removed them from the db as well (had to remove the entire instance as it was using blank nodes) and added the proper one again.

I also found an odrl:write predicate in our mu-auth config. This is equivalent to ordl:modify for mu-auth so i changed it too for good measure.

## How to test

- before running this pr, run a curl to get the dcat info: `curl -H 'Accept: text/turtle' http://localhost/dcat/catalogs > ~/Downloads/catalogs.ttl`
- apply this pr and wait for migrations to finish
- run the curl again and compare the resulting files: `curl -H 'Accept: text/turtle' http://localhost/dcat/catalogs > ~/Downloads/catalogsnew.ttl`   
The only change should be odrl:write -> odrl:modify
image


config/authorization/config.ttl
```diff
@@ -256,7 +256,7 @@ ext:allowReadForSession a odrl:Permission ;
   ext:scope "http://services.semantic.works/sessions" .
 
 ext:allowWriteForSession a odrl:Permission ;
-  odrl:action odrl:write ;
+  odrl:action odrl:modify ;
```
mirdono, line 259: Just an fyi, functionally this does not make a difference as sparql-parser's ODRL config supports `odrl:write` for reasons of backwards compatibility.  But points for consistency.

mirdono (approved):

PR #173, opened 2026-07-24
Lbron 1730 | add more filters

JonasVanHoof:
## Description

We updated the annotation review config structure, this will make current app work with that new structure

## Links to other PR's

- https://github.com/lblod/annotation-review-service/pull/21
- https://github.com/lblod/frontend-decide-human-validator/pull/28


Rahien (approved):
approving, don't forget to bump the services here as well

PR #176, opened 2026-07-27
Remove UC2 from validation dataset

brechtvdv:
Removes UC2 from dataset export by removing `schema:Answer` and `schema:Question` and `schema:Quotation` from the `human-validations.sparql` query.

# how to test

- Checkout this branch
- Run the publish dataset script: `mu script project-scripts publish-dataset --dataset human-validations --org abb` . Abb org is used, as this does not add an organization filter, so all data is outputted. When ready, you should a log `DCAT Dataset 'human-validations' written to <http://mu.semte.ch/graphs/public>.
[Pipeline] Finished. DCAT catalog + dataset written for dataset 'human-validations' (organization: 'abb') to <http://mu.semte.ch/graphs/public>.`
- Checkout the datadump that is created in the `./data/datadumps`. Use ctrl+f to search for `schema:Answer`, `schema:Quotation`, `schema:Question`. None should be found.

# Remark

Before, data from UC2 also didn't get exported, because the human validation annotation targets the question, answer, or quotation directly. There is no `?originalAnnotation` in uc2

### DIFF
--- a/scripts/project/publish_dataset/queries/human-validations.sparql
+++ b/scripts/project/publish_dataset/queries/human-validations.sparql
@@ -10,9 +10,6 @@
       ?originalAnnotation a ext:downloadResource .
       ?originalBody a ext:downloadResource .
       ?specificResource a ext:downloadResource .
-      ?quotation a ext:downloadResource .
-      ?answer a ext:downloadResource .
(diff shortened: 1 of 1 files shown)


tdn (approved):
works as intended

PR #177, opened 2026-07-28
LBRON-1709 | Cleanup previous dataset import

JonasVanHoof:
## Description

We need to cleanup the datasets in our database. We did a previous ttl insert for our own datasets but not all of them must stay in the project.

## How to test

- all not owned datasets are removed from the database
- generate the dcat-datasets from the script and see if everything turns out as expected https://github.com/lblod/app-decide/blob/development/scripts/project/publish_dataset/README.md

## Attachments

image
 

## Links to other PR's

- https://github.com/lblod/frontend-decide-dcat/pull/2
- https://github.com/lblod/dcat-service/pull/5


brechtvdv (approved):
Before: 

Image

After checking out this branch and running `drc restart migrations cache resource`:

Image

Only the two example datasets and datasets generated with the publish dataset script are retained

PR #186, opened 2026-08-04
Add shacl shape generation

brechtvdv:
This PR does following:

- generates a SHACL shape from the generated datadump ( a sh:or is used to combine all the generated node shapes). SHACL play! is used to generate the shape.
- adapt the shape URIs to make them unique per dataset uuid
- Writes the shape to the triple store
- links the DCAT dataset with the shape using dct:conformsTo
- generates a documentation page (HTML) from the adapted SHACL shape using SHACL play!
- links the DCAT dataset with the documentation page using dcat:landingPage

- The `generate_datadump_and_dcat.py` file is now split into multiple, smaller files. 
- The `lib` folder contains a generate script for datadump, dcat and shacl separately. 
- The `main.py` only tackles the CLI part. 
- A `config.py` is added for all configuration related things
- `helpers.py` is extended with sparql querying related helpers to keep generate_datadump.py minimal
- Delete previous distributions, services and landing pages linked to the dataset (so only the latest distribution and landing page is linked)

# How to test

* In `config.json`, change image to `"image": "lblod/python-semantic-works-cli:0.0.2-rc.1" (until this PR gets merged: https://github.com/lblod/python-semantic-works-cli/pull/1)

* Run script:
```
mu script project-scripts publish-dataset --dataset rmz --org abb
```
Note: a datadump_base_url must be configured in config.json in the organization specific part. Only `abb` has a default base url configured currently.

Three files should be generated in the `data/datadumps` folder:
- TIMESTAMP-rmz.ttl
- TIMESTAMP-rmz-shacl.ttl
- TIMESTAMP-rmz-shacl-adapted.ttl
- TIMESTAMP-rmz.html

* Check that the files contain relevant content.
You can check the landing page by going to the HTML file in the browser: `http://localhost/datadumps/TIMESTAMP-rmz.html`.

* Check the dataset and shacl triples in Virtuoso. The logs at the end of the run indicate the dataset URI:
`ConformsTo link for DCAT Dataset 'rmz' written to <http://mu.semte.ch/graphs/public>, combining 11 shape(s) via sh:or under <http://data.lblod.info/id/shapes/f1ad9b40-fec8-5fef-ae0e-c17a3e5abcf3>.
Execute update:
INSERT DATA { GRAPH <http://mu.semte.ch/graphs/public> {
[...]

PR #188, opened 2026-08-06
Add shapes to ldes

brechtvdv:
- Generate shapes, see: https://github.com/lblod/app-decide/pull/186
- Check LDES whether shapes are added

You should see:
image


mirdono (approved):
The shapes are published on the LDES as expected.  Have/Should we document somewhere what consumers of the feed can do with these shapes?

PR #200, opened 2026-08-12
Export AI calls properties to CSV file- #197

MartijnBogaert:
In order to get insights into cost, duration etc. of AI calls, a CSV file is generated. This results are non-aggregated by design in order for PMs to aggregate and pivot the resulting data themselves.

A mu script helps generate the CSV file, it can be triggered as follows:

```bash
mu script project-scripts export-csv --dataset ai-calls
```

By default, the CSV file is exported to the `data/csv-exports/` folder

Example output: [20260813090639-ai-calls.csv](https://github.com/user-attachments/files/31021711/20260813090639-ai-calls.csv)


scripts/project/export_csv/queries/ai-calls.sparql
```diff
@@ -0,0 +1,25 @@
...
+  ?task task:operation ?operation ;
+        ext:performedAICall ?call .
+
+  ?call ext:aiModel ?model ;
+        ext:endpoint ?endpoint ;
+        ext:tokenIn ?tokenIn ;
+        ext:tokenOut ?tokenOut .
+
+  OPTIONAL { ?call ext:duration ?durationSeconds . }
+  OPTIONAL { ?call ext:cost ?costDollar . }
```
brechtvdv: Can you add a BIND so cost is 0 as fallback?

brechtvdv (approved):

PR #201, opened 2026-08-13
LBRON-1884 | remove format as a resource and make it a property on dataset

JonasVanHoof:
## Description

As our publish dcat script says that the value of dct:format is a uri and that the frontend is not showing the format because we do not have the model in our database, we re-model it as a property of distribution

## How To Test

- application still works as expected
- together with the frontend the format is shown as a uri

## Links to other PR's

-   https://github.com/lblod/frontend-decide-dcat/pull/5
- https://github.com/lblod/app-decide-federating-catalog/pull/10

## Attachments

- http://ds.localhost/data-services will return the dataservices
image


mirdono (commented):
Left some minor comments, non-blocking.  Otherwise changes **look** goo, but need to some more testing.

mirdono (approved):

PR #204, opened 2026-08-13
Improve human validation queries

MartijnBogaert:
The human validation queries have been slightly improved:
- The codelist csv now includes a column containing the conceptscheme labels as well as the concept labels
- The NER/NEL-related csv has been renamed to `entity-recognition-validations.sparql` since entity recognition applies to all rows and entity linking only to a selection
- The NER/NEL-related csv now includes a column with the linked resources for entities where applicable
- The optionals around `?creator` have been removed since all results always have a creator


scripts/project/export_csv/queries/codelist-labeling-validations.sparql
```diff
@@ -5,14 +5,16 @@ PREFIX eli:  <http://data.europa.eu/eli/ontology#>
...
+PREFIX org:  <http://www.w3.org/ns/org#>
 
 SELECT
   ?expression
   (SAMPLE(?titel) AS ?titelBesluit)
   (SAMPLE(?datum) AS ?datumBesluit)
   ?conceptScheme
+  (SAMPLE(?conceptSchemeLabel) AS ?conceptSchemeLabel)
   ?concept
+  (SAMPLE(?conceptLabel) AS ?conceptLabel)
```
tdn: You probably know this, but in case there are multiple labels, this will just return a random one. Do we know if this is an existing problem in the data?

scripts/project/export_csv/queries/codelist-labeling-validations.sparql
tdn: was already in there, but unrelated to this PR: if `?derivedOrganizationLabel` is in this queries `GROUP BY`, can we still do a `GROUP_CONCAT` here? I'm surprised this executes at all, but it does apparently.

I do get different results locally with and without the `GROUP BY ?derivedOrganizationLabel`.

tdn (commented):
See two small comments. I think we need to inspect the actual results that come out on the test data to see if they need to be fixed or not.

scripts/project/export_csv/queries/codelist-labeling-validations.sparql
MartijnBogaert: I've checked this in the decide triplestore and there every concept/conceptscheme has only one label, that's why I _dared_ to simply sample. In the beginning, I had almost everything in the ´GROUP BY´ clause but since that made the query even slower than it currently is, I sampled wherever possible.

tdn: Alright, no problem. We're aware of it, so it should be fine.

tdn (approved):
can be merged provided the conflicts are resolved

scripts/project/export_csv/queries/codelist-labeling-validations.sparql
MartijnBogaert: Very good remark! This is something I forgot to remove from the GROUP BY statement, I removed it now.

PR #213, opened 2026-08-20
Add overview query service

brechtvdv:
# Test

checkout this branch
restart the dispatcher
drc up -d

Open in browser: `http://localhost/overview-query`, you should get a CSV with the overview

Open in browser: `http://localhost/overview-query?format=json`, you should see the JSON

### DIFF
--- a/config/overview-query/app.js
+++ b/config/overview-query/app.js
@@ -65,6 +65,7 @@ app.get("/", async function (req, res) {
         PREFIX org: <http://www.w3.org/ns/org#>
         PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
         PREFIX skos: <http://www.w3.org/2004/02/skos/core#>
+        PREFIX gold: <http://purl.org/linguistics/gold/>
 
         SELECT ?metric (str(?municipalityLabel) as ?city) ?decisions WHERE {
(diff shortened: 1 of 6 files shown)


brechtvdv:
image

config/overview-query/app.js
```diff
@@ -0,0 +1,143 @@
...
+            { ?task task:inputContainer ?c } UNION { ?task task:resultsContainer ?c } 
+            
+            # the expression itself
+            { ?c task:hasResource ?res }
+            # NER: container holds the translated expression
+            UNION { ?c task:hasResource ?x . ?res gold:translation ?x }
+            # NEL: container holds annotations on the expression
+            UNION { ?c task:hasResource ?a . ?a oa:hasTarget/oa:hasSource ?res }
+            # NEL: container holds annotations on the translated expression
+            UNION { ?c task:hasResource ?a . ?a oa:hasTarget/oa:hasSource ?y . ?res gold:translation ?y }
```
brechtvdv: This is actually wrong for what I know: annotations should point to the original expression, not the translated expression (we don't care about the translated expression)

brechtvdv: Flagged on Rocket chat, waiting for feedback from @wdmuer

brechtvdv: So this is added as fallback when the projection failed?

brechtvdv: and is this also the case with NER that we need a fallback of projection fails

Andresdev02 (approved):
Working for me. Good job @brechtvdv 👏

PR #217, opened 2026-08-26
LBRON-1916 | publish organizations dataset

JonasVanHoof:
## Description

We wan to create a dataset with all of the organizations of decide in it (Flemisch + Freiburg + Bamberg)

- create a migrations to add this dataset with its distribution
- run a construct query of all org:Organizations
- Past the output ttl in the datadump file with file name equal to the distribution accessurl
- create a query (script) for the generation of translations

## How to test

1. Run the translations & check the outcome
```
mu script project-scripts publish-dataset --dataset translations --org gent
```
2. Run the migrations so we add a dataset for organisations to the database
3. Create the dataset by just running the org file with sparql mode or executing the query and pasting the output in the correct file name
4. Check the content of the dataset (we removed AI generated + organen in tijd + bestuurseenheden that are "autonoombderijf/provincie")


config/migrations/20260826153600-add-organizations-dataset.sparql
```diff
@@ -0,0 +1,36 @@
+PREFIX dcat:  <http://www.w3.org/ns/dcat#>
+PREFIX dct:   <http://purl.org/dc/terms/>
+PREFIX foaf:  <http://xmlns.com/foaf/0.1/>
+PREFIX mu:    <http://mu.semte.ch/vocabularies/core/>
+PREFIX vcard: <http://www.w3.org/2006/vcard/ns#>
+PREFIX xsd:   <http://www.w3.org/2001/XMLSchema#>
+
+INSERT DATA {
```
Rahien: this migrations should be local, it shouldn't be run in freiburg or bamberg

config/migrations/20260826153600-add-organizations-dataset.sparql
```diff
@@ -0,0 +1,36 @@
...
+  GRAPH <http://mu.semte.ch/graphs/public> {
+    <http://data.lblod.info/id/datasets/d06c4158-b533-42d5-8fd9-ed3eb6d50e12> a dcat:Dataset;
```
Rahien: link to abb catalog using <abbcatalog> dcat:dataset <thisuri>

scripts/project/publish_dataset/config.json
```diff
@@ -51,13 +55,35 @@
...
+        "annotation"
+      ]
     },
     "human-validations": {
       "description": "Human validation annotations",
       "output_file_name": "human-validations",
       "sparql_file": "./queries/human-validations.sparql",
-      "interesting_variables": ["hvAnnotation"]
+      "interesting_variables": [
+        "hvAnnotation",
```
Rahien: i think these are only used for the count and thus we don't need more as the count will be the same?

Rahien (dismissed):
the migration you included should be a local one as we shouldn't share it with freiburg/bamberg/gent

scripts/project/publish_dataset/queries/translations.sparql
```diff
@@ -0,0 +1,36 @@
+prefix eli: <http://data.europa.eu/eli/ontology#>
+prefix oa: <http://www.w3.org/ns/oa#>
+prefix ext: <http://mu.semte.ch/vocabularies/ext/>
+prefix org: <http://www.w3.org/ns/org#>
+
+insert {
+  graph <http://mu.semte.ch/graphs/tmp-export> {
+    ?expression a ext:downloadResource .
+    ?work a ext:downloadResource .
```
Rahien: you only need the expressions in this case

config/migrations/20260826153600-add-organizations-dataset.sparql
JonasVanHoof: Added

```sparql
INSERT DATA {
  GRAPH <http://mu.semte.ch/graphs/public> {
    <http://data.lblod.info/id/catalogs/3bea809d-bfdb-485e-a7b2-10c6dea96af9> dcat:dataset <http://data.lblod.info/id/datasets/d06c4158-b533-42d5-8fd9-ed3eb6d50e12> .
  }
}
```

scripts/project/publish_dataset/queries/translations.sparql
```diff
@@ -0,0 +1,36 @@
...
+  }
+  union {
+    ?expression ext:owningBody ?organization .
+  }
+
+  optional {
+    ?expression eli:is_embodied_by ?manifests .
+  }
+  optional {
+    ?annotation oa:hasTarget ?expression .
```
brechtvdv: This only returns codelist annotations
Do you want the NER (and NEL) annotations too?

scripts/organisations-datadump.org
```diff
@@ -0,0 +1,51 @@
...
+      VALUES ?eenheidClas {
+        <http://data.vlaanderen.be/id/concept/BestuurseenheidClassificatieCode/f9cac08a-13c1-49da-9bcb-f650b0604054>
+        <http://data.vlaanderen.be/id/concept/BestuurseenheidClassificatieCode/66ec74fd-8cfc-4e16-99c6-350b35012e86>
+      }
+      ?organization besluit:bestuurt / besluit:classificatie  ?eenheidClas .
+    }
+
+    # No AI generated organizations
+    filter not exists {
+      ?statement rdf:object ?organization .
```
brechtvdv: I expected ?statement rdf:subject ?organization here 
rdf:object points to a URI that is from a registry, while the subject is the AI generated URI

brechtvdv (approved):
Maybe simplify the translations query

scripts/organisations-datadump.org
brechtvdv: Ignore my suggestion, if there is no entity linking, than it would get passed the filter, while your solution occurs on the eli:passed_by

PR #218, opened 2026-08-27
make sure we can publish datasets again if there are tombstones around for them

Rahien:
## Description
This fixes the fact that the publication script is unwilling to republish a dataset/distribution/service if a tombstone exists for its uri. These tombstones are automatically erected if such an instance is deleted by the healing of the ldes feed, but they shouldn't prevent the instance coming back from the dead.

## How to test
- take the dataset from the test environment
- use this query to remove the HVT annotations for gent:
```
    DELETE {
       GRAPH ?g {
        ?s ?p ?o.
      }
    }WHERE {
      VALUES ?s {
        <http://data.lblod.info/id/datasets/79f9ed1c-94cb-5a77-a17e-af9a192acfcb>
  <http://data.lblod.info/id/services/e433ecf7-e3c1-5c02-87d2-1829a0db838f>
  <http://data.lblod.info/id/distributions/b0687254-5c9d-553a-90f9-f7fb23e630fb>

        }
      GRAPH ?g {
        ?s ?p ?o.
      }
    }
```
- heal the ldes so the tombstones are erected: `docker compose exec ldes-delta-pusher curl -X POST http://localhost/manual-healing`
- manually add the sparql endpoint and dataset to the gent dataset in the publication config so its distribution and service are created again: 
```
      "sparql_endpoint": "http://ds.decide.lblod.info/api/sparql",
      "datadump_base_url": "http://ds.decide.lblod.info/datadumps/"
``` 
- republish the dataset: `mu script project-scripts publish-dataset --dataset human-validations --org gent`
- the dataset is now present in the dcat and the ldes

### DIFF
--- a/scripts/project/publish_dataset/lib/generate_dcat.py
+++ b/scripts/project/publish_dataset/lib/generate_dcat.py
@@ -5,15 +5,20 @@ env = Environment(loader=FileSystemLoader("."))
 from helpers import graph_has_type, delete_subjects, turtle_to_insert_data, update, log, delete_linked_resources, delete_reverse_linked_resources
 from config import datadump_file_name, dataset_uri_and_uuid, distribution_uri_and_uuid, service_uri_and_uuid, PUBLIC_GRAPH
 
+DCAT_CATALOG = "http://www.w3.org/ns/dcat#Catalog"
+DCAT_DATASET = "http://www.w3.org/ns/dcat#Dataset"
+DCAT_DISTRIBUTION = "http://www.w3.org/ns/dcat#Distribution"
+DCAT_DATASERVICE = "http://www.w3.org/ns/dcat#DataService"
+
 def step1_write_catalog(organization: str, organization_config: dict, now_iso: str) -> None:
     log("[Step 1] Generate DCAT Catalog for %s", organization)
     catalog_uri = organization_config["catalog_uri"]
     catalog_publisher_uri = organization_config["catalog_publisher"].get("uri")
     
-    if graph_has_type(catalog_uri, "http://www.w3.org/ns/dcat#Catalog", PUBLIC_GRAPH):
+    if graph_has_type(catalog_uri, DCAT_CATALOG, PUBLIC_GRAPH):
         log("Catalog '%s' already exists in <%s>, doing nothing.", organization, PUBLIC_GRAPH)
     else:
-        delete_subjects([catalog_uri]) # in case there is a tombstone for it
+        delete_subjects([catalog_uri], PUBLIC_GRAPH) # in case there is a tombstone for it or any other information that claims it is some other type than a catalog
         catalog_template = env.get_template("templates/dcat-catalog.ttl.j2")
         catalog_output = catalog_template.render(
             ISSUED=now_iso,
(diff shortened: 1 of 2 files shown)


scripts/project/publish_dataset/lib/generate_dcat.py
```diff
@@ -2,18 +2,18 @@
...
 
 def step1_write_catalog(organization: str, organization_config: dict, now_iso: str) -> None:
     log("[Step 1] Generate DCAT Catalog for %s", organization)
     catalog_uri = organization_config["catalog_uri"]
     catalog_publisher_uri = organization_config["catalog_publisher"].get("uri")
-    catalog_subjects = [catalog_uri] + ([catalog_publisher_uri] if catalog_publisher_uri else [])
-
-    if graph_has_subject(catalog_uri, PUBLIC_GRAPH):
+    
+    if graph_has_type(catalog_uri, "http://www.w3.org/ns/dcat#Catalog", PUBLIC_GRAPH):
```
mirdono: Nitpick: For readability I would extract the type to a constant.

Imo that would make the code slightly easier to read.  Similar for other types used throughout this module.

scripts/project/publish_dataset/lib/generate_dcat.py
```diff
@@ -2,18 +2,18 @@
...
         log("Catalog '%s' already exists in <%s>, doing nothing.", organization, PUBLIC_GRAPH)
     else:
+        delete_subjects([catalog_uri]) # in case there is a tombstone for it
```
mirdono: Issue: Should also provide the graph as argument.

This function also requires the graph as argument, otherwise its contained query will not be correct.  To test I deleted the relevant catalogue resource and tried to regenerate the same dataset as in your instructions.  This resulting in the following error:

``` shell
Traceback (most recent call last):
  File "/script/./main.py", line 51, in <module>
    generate_dcat(now, args.dataset, dataset_config, args.org, organization_config)
    ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/script/lib/generate_dcat.py", line 95, in generate_dcat
    step1_write_catalog(organization, organization_config, timestamp_iso)
    ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/script/lib/generate_dcat.py", line 16, in step1_write_catalog
    delete_subjects([catalog_uri]) # in case there is a tombstone for it
    ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
TypeError: delete_subjects() missing 1 required positional argument: 'graph'
```

Furthermore, the comment on this line suggests the deletion is conditional on the existence of a tombstone.  But I fail to see any functionality actually checking for one.  Are we at risk here of deleting too much data?

mirdono (changes requested):
Ran into an issue when it tries to delete an existing/tombstoned catalogue.  See inline comment for details.

scripts/project/publish_dataset/lib/generate_dcat.py
Rahien: Fixed the missing argument, good catch! For the possibility of deleting too much data: yes i guess there is a risk. However, this only happens if the subject exists without the type dcat:Catalog in our own database. This feels unlikely except for the case of it being deleted and a tombstone being left behind. Also, this script is built very specifically for DECIDe, I wouldn't reuse in other projects.

mirdono (approved):

